PR #20, opened 2018-09-10
Fix SignalRConnectionInfo to support negotiate

chenkennt:
Rename properties in SignalRConnectionInfo to comply with SignalR protocol.

### DIFF
--- a/samples/simple-chat/content/index.html
+++ b/samples/simple-chat/content/index.html
@@ -25,27 +25,23 @@
   <script src="https://unpkg.com/@aspnet/signalr@1.0/dist/browser/signalr.js"></script>
   <script src="https://unpkg.com/axios/dist/axios.min.js"></script>
   <script>
-    const apiBaseUrl = 'http://localhost:7071';
-    const hubName = 'simplechat';
+    const apiBaseUrl = 'http://localhost:7071/simplechat';
 
-    getConnectionInfo().then(info => {      
+    (function () {
       let username;
       while (!username && username !== null) {
         username = prompt('Enter a username');
       }
-      
+
       if (username === null) return;
       document.body.classList.add('ready');
 
       const messageForm = document.getElementById('message-form');
       const messageBox = document.getElementById('message-box');
       const messages = document.getElementById('messages');
-      const options = {
-        accessTokenFactory: () => info.accessKey
-      };
 
       const connection = new signalR.HubConnectionBuilder()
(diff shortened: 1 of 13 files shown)


chenkennt:
Fixed

anthonychu:
@chenkennt I don't think the updated client-side JavaScript will work, as CORS in Azure App Service and also the Functions runtime locally have a problem when the request is made with credentials.

https://github.com/Azure/azure-functions-host/issues/620

I recommend changing back to making the negotiate request manually until we can think of a better solution.

chenkennt:
CORS should be a different problem. I would recommend to use function proxy so that both client page and REST API are in the same domain. I'm wondering how does the binding work right now? It's a cross origin request anyway even without this change.

anthonychu:
Not everyone can use proxies to serve their app, and we're recommending static websites in Storage for serverless scenarios. Proxies also require extra steps in tutorials which add to complexity.

This is only a problem when the HTTP request is made with credentials. The simple chat sample works because it is anonymous. The chat with auth example in #22 (and all cross domain EasyAuth examples currently) works because the auth token is passed via a non-standard header `x-zumo-auth`; if it were `Authorization`, it would have necessitated making the call with credentials and it would fail on App Service / Functions CORS.

It was my intention to make this work seamlessly as well, which is why the object was modelled after the negotiate response (except I got the field names wrong 😃). I think we should ship the samples with the explicit call, then work with app service team to fix this.

We can also look at whether we can ship a JS shim to get around this in the meantime that allows the client to make the negotiate call without credentials and pass a custom header like `x-zumo-auth`.

chenkennt:
According to workaround provided in the issue, I'd like to add the workaround to the function code:
```js
  context.res = {
    body: connectionInfo,
    headers: {
      'Access-Control-Allow-Credentials': 'true',
      'Access-Control-Allow-Origin': origin,
      'Access-Control-Allow-Methods': 'GET, OPTIONS',
      'Access-Control-Allow-Headers': 'X-Requested-With'
    }
  };
```
What do you think?

anthonychu:
Sounds good. Please test locally and in Azure to confirm the workaround works. Thanks!

chenkennt:
Added CORS header to functions. To do this I have to change your simplified C# sample back to parse JSON from body.

anthonychu:
Looks good. Let's merge and try to publish to NuGet. Thanks.

PR #22, opened 2018-09-10
Add support for users

### LINKED ISSUE #14: Add support for users
Planning to add support for users in the following manner...

## `SignalRConnectionInfo` input binding

Add a `userId` property to the binding. If using EasyAuth, it can be set to the user id (a GUID) or the username (email or handle, depending on provider) using a binding expression

```javascript
{  
  "type": "signalRConnectionInfo",  
  "name": "connectionInfo",  
  "hubName": "simplechat",  
  "userId": "{headers.x-ms-client-principal-id}", 
  "direction": "in" 
}
```

> SignalR uses the `System.Security.Claims.ClaimTypes.NamedIdentifier` claim to determine the user id of the connection.

## `SignalR` output binding

Add a `UserId` property to the `SignalRMessage` class to indicate a user id to receive a message. If this property is set, the message is sent to the user instead of broadcasting it to all connections.

### Updated `SignalRMessage` object

```javascript
{
  "userId": "userId1",
  "target": "targetMethodName",
  "arguments": [ "arg1", "arg2" ]
}
```

### REST API

Call `/api/v1-preview/hub/{hub}/user/{user}`
https://github.com/Azure/azure-signalr/blob/dev/docs/swagger/v1-preview.json

anthonychu:
See #14 
Closes #14


xscript:
@anthonychu Send to users will be removed in the version `v1` of REST API. Do we still want to support it?

anthonychu:
It will? Perhaps we should change it to take a single user id per message. I can make that change. Do you agree?

PR #24, opened 2018-09-14
update build to use KoreBuild

chenkennt:


Directory.Build.props
```diff
@@ -0,0 +1,26 @@
+﻿<Project>
+  <Import
```
xscript: This can be removed since `AspNetCoreSettings` is not related here.

Directory.Build.props
```diff
@@ -0,0 +1,26 @@
...
+    Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), AspNetCoreSettings.props))\AspNetCoreSettings.props"
+    Condition=" '$(CI)' != 'true' AND '$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), AspNetCoreSettings.props))' != '' " />
+
+  <Import Project="version.props" />
+  <Import Project="build\dependencies.props" />
+  <Import Project="build\sources.props" />
+
+  <PropertyGroup>
```
xscript: Need to set `projectIconUrl`, otherwise the package can't be published to NuGet.

build/dependencies.props
```diff
@@ -0,0 +1,16 @@
+<Project>
+  <PropertyGroup>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+    <RepositoryRoot>$(MSBuildThisFileDirectory)\..</RepositoryRoot>
+  </PropertyGroup>
+  <PropertyGroup Label="Package Versions">
+    <MicrosoftAzureWebJobsPackageVersion>3.0.0-beta8</MicrosoftAzureWebJobsPackageVersion>
```
xscript: nit: Better to sort them in alphabetic order for easier maintenance.

global.json
```diff
@@ -0,0 +1,5 @@
+{ "sdk": {
```
xscript: nit: format

chenkennt: Looks like it's regenerated every time a build is triggered so I'll just leave it as is.

PR #31, opened 2018-10-15
Update to use API v1

### LINKED ISSUE #28: Update to v1.0 of the service REST API
* Change to port 443
* Update endpoints https://github.com/Azure/azure-signalr/blob/dev/docs/rest-api.md

@chenkennt How should we use the version number that is now in the connection string?

anthonychu:
Closes #28


mahpour:
When is this expected to be available in the portal?

PR #53, opened 2019-03-04
Add constructors to SignalRMessage and SignalRGroupAction

anthonychu:
Make it easier to instantiate these objects properly.

### DIFF
--- a/binding-library/java/src/main/java/com/microsoft/azure/functions/signalr/SignalRMessage.java
+++ b/binding-library/java/src/main/java/com/microsoft/azure/functions/signalr/SignalRMessage.java
@@ -16,6 +16,26 @@ import java.util.List;
  * @since 1.0.0
  */
 public class SignalRMessage {
+
+    /**
+     * Constructor
+     */
+    public SignalRMessage() {
+    }
+
+    /**
+     * Constructor
+     * 
+     * @param target    Target method to invoke on clients
+     * @param arguments Arguments to pass to target method
+     */
+    public SignalRMessage(String target, Object... arguments) {
+        this.target = target;
+        for (Object arg: arguments) {
+            this.arguments.add(arg);
+        }
+    }
+
     /**
      * User to send the message to
      */
--- a/binding-library/java/src/main/java/com/microsoft/azure/functions/signalr/SignalRMessage.java
+++ b/binding-library/java/src/main/java/com/microsoft/azure/functions/signalr/SignalRMessage.java
@@ -6,6 +6,7 @@
 package com.microsoft.azure.functions.signalr;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


binding-library/java/src/main/java/com/microsoft/azure/functions/signalr/SignalRMessage.java
```diff
@@ -16,6 +16,26 @@
...
+     * Constructor
+     * 
+     * @param target    Target method to invoke on clients
+     * @param arguments Arguments to pass to target method
+     */
+    public SignalRMessage(String target, Object... arguments) {
+        this.target = target;
+        for (Object arg: arguments) {
+            this.arguments.add(arg);
+        }
```
JialinXin: this.arguments.add(arguments) directly?

binding-library/java/src/main/java/com/microsoft/azure/functions/signalr/SignalRMessage.java
```diff
@@ -16,6 +16,26 @@
...
+    public SignalRMessage() {
+    }
+
+    /**
+     * Constructor
+     * 
+     * @param target    Target method to invoke on clients
+     * @param arguments Arguments to pass to target method
+     */
+    public SignalRMessage(String target, Object... arguments) {
```
JialinXin: >public SignalRMessage( [](start = 4, length = 22)

Shall we also add one contractor with userId that for send to user and one contractor with groupname that for send to group?

JialinXin: That should be handle in function. Ignore this.

---
In reply to: [261928906](https://github.com/Azure/azure-functions-signalrservice-extension/pull/53#discussion_r261928906) [](ancestors = 261928906)

anthonychu: Been thinking about this one. Because they are both String, I'm not sure how we can differentiate. Potentially, we can remove userId and groupName from `SignalRMessage` and create subclasses `SignalRGroupMessage` and `SignalRUserMessage` that have those as fields and in the constructors. This is probably a better way to do it, but might be too late to make this big of a change.

JialinXin: Yes. I'm thinking if it should be handled in app level or function level. It's not block issue. Let's consider it in next releasing cycle. 

---
In reply to: [261930864](https://github.com/Azure/azure-functions-signalrservice-extension/pull/53#discussion_r261930864) [](ancestors = 261930864)

binding-library/java/src/main/java/com/microsoft/azure/functions/signalr/SignalRMessage.java
```diff
@@ -16,6 +17,24 @@
...
+        this.target = target;
+        this.arguments.addAll(Arrays.asList(arguments));
```
JialinXin: 👍 That's what I'm looking for!

PR #55, opened 2019-03-26
README.md Update - Removed the groups point under Current Limitations

tehmas:


JialinXin:
Thank you @tehmas for contributing! Currently, we support broadcasting, send to user, send to group, add/remove user in group. But still not support targeting a subset of connections, users, or groups. Suppose we should update the description instead of remove it.

PR #57, opened 2019-04-24
Wanl/use mng sdk

realwanpengli:
replace HttpClient by ServiceHubContext, which can use both rest api/websockets connection to send messages.


src/SignalRServiceExtension/Client/AzureSignalRClient.cs
```diff
@@ -4,88 +4,64 @@
...
-            var hubUrl = $"{BaseEndpoint}/api/v1/hubs/{hubName}";
-            var audienceUrl = $"{hubUrl}{additionalPath}";
-            var token = GenerateJwtBearer(null, audienceUrl, null, DateTime.UtcNow.AddMinutes(30), AccessKey);
-            return new SignalRConnectionInfo
-            {
-                Url = audienceUrl,
-                AccessToken = token
-            };
+            var serviceHubContext = await serviceHubContextStore.GetOrAddAsync(hubName);
+            await serviceHubContext.Clients.All.SendCoreAsync(data.Target, data.Arguments);
```
JialinXin: >SendCoreAsync [](start = 48, length = 13)

When migration to ManagementSDK, make sure to add ProductInfo correctly for the request.

src/SignalRServiceExtension/Config/ServiceHubContextStore.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.Azure.SignalR.Management;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    internal class ServiceHubContextStore : IServiceHubContextStore
+    {
+        private readonly ConcurrentDictionary<string, (Lazy<Task<IServiceHubContext>> lazy, IServiceHubContext value)> store = new ConcurrentDictionary<string, (Lazy<Task<IServiceHubContext>>, IServiceHubContext value)>();
+        private readonly IServiceManager serviceManager;
+        private readonly ILoggerFactory loggerFactory;
```
JialinXin: >loggerFactory [](start = 40, length = 13)

Is loggerFactory in use?

src/SignalRServiceExtension/Config/SignalROptions.cs
```diff
@@ -1,11 +1,14 @@
...
 
+using Microsoft.Azure.SignalR.Management;
+
 namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
     public class SignalROptions
     {
         public string ConnectionString { get; set; }
         public string HubName { get; set; }
+        public ServiceTransportType ServiceTransportType { get; set; } = ServiceTransportType.Persistent;
```
JialinXin: >Persistent [](start = 94, length = 10)

Shall we use Transient mode as default? Suppose user working on Azure Function would be familar with RestApi and it doesn't have any requirement on function settings, like always on.
Also update docs to explain this configuration.

build/dependencies.props
```diff
@@ -8,9 +8,9 @@
     <MicrosoftAzureWebJobsPackageVersion>3.0.4</MicrosoftAzureWebJobsPackageVersion>
     <MicrosoftNETTestSdkPackageVersion>15.8.0</MicrosoftNETTestSdkPackageVersion>
     <MoqPackageVersion>4.9.0</MoqPackageVersion>
-    <SystemIdentityModelTokensJwtPackageVersion>5.1.4</SystemIdentityModelTokensJwtPackageVersion>
     <XunitPackageVersion>2.4.0</XunitPackageVersion>
     <XunitRunnerVisualstudioPackageVersion>2.4.0</XunitRunnerVisualstudioPackageVersion>
+    <MicrosoftAzureSignalRManagement>1.0.0-preview1-10351</MicrosoftAzureSignalRManagement>
```
JialinXin: move up in order.

samples/simple-chat/csharp/FunctionApp/local.settings.sample.json
```diff
@@ -4,7 +4,8 @@
     "AzureWebJobsStorage": "",
     "AzureWebJobsDashboard": "",
     "FUNCTIONS_WORKER_RUNTIME": "dotnet",
-    "AzureSignalRConnectionString": "<signalr-connection-string>"
+    "AzureSignalRConnectionString": "<signalr-connection-string>",
+    "ServiceTransportType": "Persistent"
```
JialinXin: >ServiceTransportType [](start = 5, length = 20)

rename to AzureSignalRServiceTransportType

src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
```diff
@@ -1,39 +1,45 @@
...
+using System.Threading.Tasks;
 
 namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
     [Extension("SignalR")]
     internal class SignalRConfigProvider : IExtensionConfigProvider
     {
         internal const string AzureSignalRConnectionStringName = "AzureSignalRConnectionString";
-
+        private const string ServiceTransportTypeName = "ServiceTransportType";
```
JialinXin: >ServiceTransportType [](start = 57, length = 20)

rename to AzureSignalRServiceTransportType for consistency

src/SignalRServiceExtension/Config/ServiceHubContextStore.cs
realwanpengli: It is used in line 26 in this file.

JialinXin: OK. I didn't see this.

src/SignalRServiceExtension/Client/AzureSignalRClient.cs
realwanpengli: If ServiceTransportType is Transient, the ProductInfo can be added to header,
but if the ServiceTransportType is Persistent, messages will be sent in websockets connection. There seems no way to declare `Asrs-User-Agent` like REST API. 
> request.Headers.Add("Asrs-User-Agent", GetProductInfo())

Is it OK that only Transient mode adds the Product Info?

JialinXin: When use websockets connection we doesn't log message send requests. But could it be added when setup connection between functionapp and signalr service similar to the behavior between normal app server and signalr service. So it could be tracked.

realwanpengli: I reuse the signalr SDK, therefore, in Persistent mode, the product info is already added when starting connection to signalr service, but the package name will be `management SDK` since signalr sdk calls `Assembly.GetCallingAssembly()` to obtain package name. 
Although we can change the signalr sdk's interface to pass the assembly name, it is not elegant...

realwanpengli: Yes. Since Persistent mode reuse service sdk to start connection, the production info is already in request header. But the problem is the package name will be Management SDK instead of . I need to change the api of server-connection-start-method.

realwanpengli: Will be add in management sdk

src/SignalRServiceExtension/Config/SignalROptions.cs
realwanpengli: Sounds good. What's more, the deafult mode in management sdk is transient.

PR #59, opened 2019-05-01
Add SignalRIAsyncCollectorExtensions to simulate an IHubContext in Azure Functions

halter73:
An IHubContext is a little nicer and more familiar than using an IAsyncCollector directly so I created the GetHubClients() and GetHubGroups() extension methods for `IAsyncCollector<SignalRMessage>` and `IAsyncCollector<SignalRGroupAction>` respectively to simulate an IHubContext.

Before:

```C#
await collector.AddAsync(new SignalRMessage
{
    Target = "newMessage",
    Arguments = new[] { "arg1", "arg2" }
});
await collector.AddAsync(new SignalRMessage
{
    UserId = "userId1",
    Target = "newMessage",
    Arguments = new[] { "arg1", "arg2" }
});
await collector.AddAsync(new SignalRMessage
{
    GroupName = "group1",
    Target = "newMessage",
    Arguments = new[] { "arg1", "arg2" }
});
```

After:

```C#
var clients = collector.GetHubClients();
await clients.All.SendCoreAsync("newMessage", new[] { "arg1", "arg2" });
await clients.User("userId1").SendCoreAsync("newMessage", new[] { "arg1", "arg2" });
await clients.Group("group1").SendCoreAsync("newMessage", new[] { "arg1", "arg2" });
```

IHubClients.Client, IHubClients.Clients, IHubClients.AllExcept, and IHubClients.GroupExcept all throw NotImplementedExceptions, but these could all be trivially added once it's supported by SignalRMessage.

Long term it might be nice to unify SignalRMessage and SignalRGroupAction so you could create a full IHubContext from a single IAsyncCollector.

### DIFF
--- a/src/SignalRServiceExtension/Bindings/SignalRIAsyncCollectorExtensions.cs
+++ b/src/SignalRServiceExtension/Bindings/SignalRIAsyncCollectorExtensions.cs
@@ -1,0 +1,152 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Azure.SignalR.Management;
+using Microsoft.Azure.WebJobs.Extensions.SignalRService;
+
+namespace Microsoft.Azure.WebJobs
+{
+    public static class SignalRIAsyncCollectorExtensions
+    {
+        public static IHubClients GetHubClients(this IAsyncCollector<SignalRMessage> collector)
+        {
+            return new CollectorHubClients(collector);
+        }
+
+        public static IUserGroupManager GetHubGroups(this IAsyncCollector<SignalRGroupAction> collector)
+        {
+            return new CollectorUserGroupManager(collector);
+        }
+
+        private class CollectorClientProxy : IClientProxy
+        {
+            private readonly IAsyncCollector<SignalRMessage> _collector;
+            private IEnumerable<SignalRMessage> _destinations;
+
+            public CollectorClientProxy(IAsyncCollector<SignalRMessage> collector, SignalRMessage destination)
+                : this(collector, new[] { destination })
+            {
+            }
+
+            public CollectorClientProxy(IAsyncCollector<SignalRMessage> collector, IEnumerable<SignalRMessage> destinations)
+            {
+                _collector = collector;
+                _destinations = destinations;
+            }
+
+            public async Task SendCoreAsync(string method, object[] args, CancellationToken cancellationToken = default)
+            {
+                foreach (var destination in _destinations)
+                {
+                    await _collector.AddAsync(new SignalRMessage
+                    {
+                        GroupName = destination.GroupName,
+                        UserId = destination.UserId,
+                        Target = method,
+                        Arguments = args,
+                    }, cancellationToken);
+                }
+            }
+        }
+
+        private class CollectorHubClients : IHubClients
+        {
+            private readonly IAsyncCollector<SignalRMessage> _collector;
+
+            public CollectorHubClients(IAsyncCollector<SignalRMessage> collector)
+            {
+                _collector = collector;
+            }
+
+            public IClientProxy All => new CollectorClientProxy(_collector, new SignalRMessage());
+
+            public IClientProxy AllExcept(IReadOnlyList<string> excludedConnectionIds)
+            {
+                throw new NotImplementedException();
+            }
+
+            public IClientProxy Client(string connectionId)
+            {
+                throw new NotImplementedException();
+            }
+
+            public IClientProxy Clients(IReadOnlyList<string> connectionIds)
+            {
+                throw new NotImplementedException();
+            }
+
+            public IClientProxy Group(string groupName)
+            {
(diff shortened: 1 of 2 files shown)


halter73:
@davidfowl Our talk got me looking more into our serverless stuff, and while this isn't the server-client-hybrid thing we discussed, this does give you an IHubContext like thing in Azure functions.

realwanpengli:
Hi @halter73 , thanks for contributing! The `SignalRAsyncCollector` uses `ServiceHubContext` for each hub to send messages. The `ServiceHubContext` is a `IHubContext` like thing (with `IUserGroupManager`).

If we can expose the GetOrAddAsync() method, we can directly use the `ServiceHubContext`. How about it?

JialinXin:
Thank you @halter73 for taking a look into this! It's more straight for csharp developers in this way. I did some tests on this change and have a little concern. 
1. Do you know how it works for non .NET developers like javascript?
2. Under this implementation, users have to detect send behavior by parameters, either split to different functions or did some pre-check then call target methods. We asked users to split the methods but then we build them to a generic `SignalRMessage` in `SendAsyncCore()` and call `AddSync()`, finally it needs to check parameters and call different methods in `AzureSignalRClient` again. It seems complicate the behavior.

halter73:
1. This change only affects .NET developers. These extension methods only build a more familiar API on top of `IAsyncCollector<SignalRMessage>` and `IAsyncCollector<SignalRGroupAction>`.

2. I'm not sure I understand the second question. This change is only additive and should be completely non-breaking. I don't know what it means for users to have to "detect send behavior by parameters". Obviously, anything you can do via the IHubClients returned by GetHubClients() or the IUserGroupManager returned by GetHubGroups() could also be done by calling AddAsync() directly. If users prefer AddAsync(), nothing prevents them from continuing to do so.

As for exposing a ServiceHubContext directly from the SignalRAsyncCollector, I assumed that wasn't directly possible. With all the serialization and deserialization of messages going on, I assumed that `IAsyncCollector<SignalRMessage>` passed to the Azure Function is actually a different type that serializes the messages to be passed to the actual SignalRAsyncCollector. Is this not the case?

JialinXin:
User can use same method for different send message purpose in the old way. And in the new way, they could be more straight if they use specific destinations. You're right, it's not breaking and users still have the choice. Good to have this. Thank you!

src/SignalRServiceExtension/Bindings/SignalRIAsyncCollectorExtensions.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Azure.SignalR.Management;
+using Microsoft.Azure.WebJobs.Extensions.SignalRService;
+
+namespace Microsoft.Azure.WebJobs
+{
+    public static class SignalRIAsyncCollectorExtensions
```
JialinXin: Extra `I`?

src/SignalRServiceExtension/Bindings/SignalRIAsyncCollectorExtensions.cs
```diff
@@ -0,0 +1,152 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Azure.SignalR.Management;
+using Microsoft.Azure.WebJobs.Extensions.SignalRService;
+
+namespace Microsoft.Azure.WebJobs
```
JialinXin: use namespace `Microsoft.Azure.WebJobs.Extensions.SignalRService` and could remove the using reference

src/SignalRServiceExtension/Bindings/SignalRIAsyncCollectorExtensions.cs
halter73: I don't think so. This class contains extension methods for `IAsyncCollector`.

src/SignalRServiceExtension/Bindings/SignalRIAsyncCollectorExtensions.cs
halter73: [IAsyncCollector](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs/IAsyncCollector.cs) lives in Microsoft.Azure.WebJobs. Generally we like to keep extension methods in the same namespace as the class they're extending so the extension methods always show up in intellisense.

src/SignalRServiceExtension/Bindings/SignalRIAsyncCollectorExtensions.cs
JialinXin: Oh, you consider it extension of `IAsyncCollector` instead of `SignalRAsyncCollector`. Would it be extend SignalRAsyncCollector more accurate?

realwanpengli:
> As for exposing a ServiceHubContext directly from the SignalRAsyncCollector, I assumed that wasn't directly possible.
@halter73 I think exposing a `IServiceHubContext` is possible, please take a look at PR #62 .
Interfaces of your extensions is already available in Management SDK. What's more, Management SDK is adding more features to `IServieHubContext`. Exposing `IServieHubContext` is simple, and reduces future work to extend it.

halter73:
Closing since #62 makes this redundant.

PR #60, opened 2019-05-05
Update to the correct version for the extension

dfinke:


anthonychu:
Thanks for catching this. Sorry it sent you down the wrong path.

PR #62, opened 2019-05-09
Exposing servicehubcontext for improving user experience 

realwanpengli:
Exposing `IServiceHubContext` for C# users. This interface is an extened interface of `IHubContext`.

Alternative for #59


src/SignalRServiceExtension/Bindings/SignalRIAsyncCollectorExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.SignalR.Management;
+using Microsoft.Azure.WebJobs.Extensions.SignalRService;
+
+namespace Microsoft.Azure.WebJobs
+{
+    public static class SignalRIAsyncCollectorExtensions
+    {
+        public static ValueTask<IServiceHubContext> GetServiceHubContextAsync(this IAsyncCollector<SignalRMessage> collector, string hubName) =>
```
JialinXin: Add one for `IAsyncCollector<SignalRGroupAction>` as well?

halter73: It's cool that this works. I definitely prefer a fully-functional IServiceHubContext over my PR.

I do think it's weird to make this an extension method on `IAsyncCollector<SignalRMessage>` considering GetServiceHubContextAsync doesn't even use the collector.

I would probably just add a public static `ServiceHubContextStore.GetServiceHubContextAsync` method that does the same thing and calls StaticServiceHubContextStore.ServiceHubContextStore.GetOrAddAsync(hubName).

realwanpengli: I think a generic type `IAsyncCollector<T>.GetServiceHubContextAsync(...)` is better, since extension methods show up in intellisense is good like you said.

halter73: It would be nice to be able to use the IServiceHubContext even if you didn't take an `IAsyncCollector<T>` parameter. I could see an argument for having both an extension method and normal static method.

halter73:
One thing I'm really curious about is what's the purpose of types like [SignalROutputConverter](https://github.com/Azure/azure-functions-signalrservice-extension/blob/dev/src/SignalRServiceExtension/Bindings/SignalROutputConverter.cs) which converts JObjects to SignalRMessages and SignalRGroupActions and is called [here in SignalRAsyncCollector.AddAsync()](https://github.com/Azure/azure-functions-signalrservice-extension/blob/304585fd89c0f9db480163080eb419e322d7e69a/src/SignalRServiceExtension/Bindings/SignalRAsyncCollector.cs#L30) if you can use a IServiceHubContext directly inside an Azure Function without converting SignalRMessages to JObjects and back.

It seems really wasteful to do that unless you *have* to marshal the SignalRMessages to some other domain/context to actually use the IServiceHubContext. But if that were the case, this PR would not work. Is the marshalling done just to support other languages or while being unnecessary in C#? Or is there some other reason for the marshalling?

JialinXin:
It's used for other language mainly for javascript. C# and java doesn't have this problem cause they know what it is when call AddAsync(), but javascript always make it JObjects. Javascript is the most used language in Azure Functions and that's why we take them into consideration when adding new features.

src/SignalRServiceExtension/Bindings/SignalRIAsyncCollectorExtensions.cs
realwanpengli: Sounds good. C# users don't need to use the binding `SignalRAttribute` if they like.

halter73:
I wanted to try this out myself in a real Functions app instead of just inside of test methods, but I was unable to follow the quickstart at https://docs.microsoft.com/en-us/azure/azure-signalr/signalr-quickstart-azure-functions-csharp

Can you please take a look at https://github.com/Azure-Samples/signalr-service-quickstart-serverless-chat/issues/9? Thanks!

JialinXin:
@halter73 I'm not able to repro the problem. I've replied a few in the issue and meanwhile have you tried the sample(https://github.com/Azure/azure-functions-signalrservice-extension/tree/dev/samples/simple-chat/csharp/FunctionApp) from current repo? They should be same and easy for debugging.

PR #68, opened 2019-07-11
Add connection id related apis

zackliu:


src/SignalRServiceExtension/Bindings/SignalRAsyncCollector.cs
```diff
@@ -43,7 +43,11 @@ public async Task AddAsync(T item, CancellationToken cancellationToken = default
                     throw new ArgumentException("GroupName and UserId can not be specified at the same time.");
                 }
```
JialinXin: since there's one more parameter added and have priority order, either do the check more accurately or leave it with priority logic in below but give users guidance in some docs.

src/SignalRServiceExtension/Client/AzureSignalRClient.cs
```diff
@@ -60,6 +60,12 @@ public async Task SendToAll(string hubName, SignalRData data)
             await serviceHubContext.Clients.All.SendCoreAsync(data.Target, data.Arguments);
         }
 
+        public async Task SendToConnection(string hubName, string connectionId, SignalRData data)
+        {
+            var serviceHubContext = await serviceHubContextStore.GetOrAddAsync(hubName);
+            await serviceHubContext.Clients.Client(connectionId).SendCoreAsync(data.Target, data.Arguments);
```
realwanpengli: In Management SDK, only Persistent mode support it for now.

src/SignalRServiceExtension/Client/AzureSignalRClient.cs
```diff
@@ -108,6 +114,34 @@ public async Task RemoveUserFromGroup(string hubName, string userId, string grou
...
+            if (string.IsNullOrEmpty(connectionId))
+            {
+                throw new ArgumentException($"{nameof(connectionId)} cannot be null or empty");
+            }
+            if (string.IsNullOrEmpty(groupName))
+            {
+                throw new ArgumentException($"{nameof(groupName)} cannot be null or empty");
+            }
+            var serviceHubContext = await serviceHubContextStore.GetOrAddAsync(hubName);
+            await serviceHubContext.Groups.AddToGroupAsync(connectionId, groupName);
```
realwanpengli: In Management SDK, only Persistent mode support it for now.

src/SignalRServiceExtension/Client/AzureSignalRClient.cs
```diff
@@ -108,6 +114,34 @@ public async Task RemoveUserFromGroup(string hubName, string userId, string grou
...
+            if (string.IsNullOrEmpty(connectionId))
+            {
+                throw new ArgumentException($"{nameof(connectionId)} cannot be null or empty");
+            }
+            if (string.IsNullOrEmpty(groupName))
+            {
+                throw new ArgumentException($"{nameof(groupName)} cannot be null or empty");
+            }
+            var serviceHubContext = await serviceHubContextStore.GetOrAddAsync(hubName);
+            await serviceHubContext.Groups.RemoveFromGroupAsync(connectionId, groupName);
```
realwanpengli: In Management SDK, only Persistent mode supports it for now.

src/SignalRServiceExtension/Bindings/SignalRAsyncCollector.cs
```diff
@@ -59,13 +63,28 @@ public async Task AddAsync(T item, CancellationToken cancellationToken = default
...
-                    await client.RemoveUserFromGroup(hubName, groupAction.UserId, groupAction.GroupName).ConfigureAwait(false);
+                    if (groupAction.Action == GroupAction.Add)
+                    {
+                        await client.AddUserToGroup(hubName, groupAction.UserId, groupAction.GroupName).ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        await client.RemoveUserFromGroup(hubName, groupAction.UserId, groupAction.GroupName).ConfigureAwait(false);
+                    }
                 }
```
realwanpengli: If both user ID and connection ID are null or empty. The exception will always be `ArgumentException($"{nameof(userId)} cannot be null or empty");` but we don't know which one is actually missing.

src/SignalRServiceExtension/Bindings/SignalRMessage.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
     [JsonObject]
     public class SignalRMessage
```
JialinXin: >   public class SignalRMessage [](start = 1, length = 30)

remember to support Java as well.

src/SignalRServiceExtension/Client/AzureSignalRClient.cs
zackliu: I will update Management SDK first

src/SignalRServiceExtension/Bindings/SignalRAsyncCollector.cs
zackliu: Discussed with jialin, I want to use a priority order to resolve the parameters. `ConnectionId` -> `UserId` -> `GroupName`. Thus, customer can set both `ConnectionId` and `UserId` and we can update the exception message to something like `connectionId and userId cannot be null or empty at the same time.`

PR #80, opened 2019-10-12
A POC on SignalR Trigger

zackliu:
* Allow sending messages through client to Function
* Invoking by FunctionName just like hub with server
* TODO: support true invokeAsync


src/SignalRServiceExtension/TriggerBindings/SignalRTriggerRouter.cs
```diff
@@ -0,0 +1,112 @@
...
+                        invocationContext = JsonConvert.DeserializeObject<InvocationContext>(body);
+                    }
+                    catch (Exception)
+                    {
+                        return new HttpResponseMessage(HttpStatusCode.BadRequest);
+                    }
+
+                    if (string.IsNullOrEmpty(invocationContext.Data?.Target))
+                    {
+                        return new HttpResponseMessage(HttpStatusCode.BadRequest);
```
realwanpengli: More info about what argument is bad?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerRouter.cs
```diff
@@ -0,0 +1,112 @@
...
+                    {
+                        invocationContext.HubName = hubName;
+                        var signalRTriggerEvent = new SignalRTriggerEvent
+                        {
+                            Context = invocationContext,
+                        };
+
+                        // TODO: select out listener that match the pattern
+
+                        var result = await listener.Executor.TryExecuteAsync(new Host.Executors.TriggeredFunctionData
```
realwanpengli: Looks like we can only store `Executor` instead of the whole `listener`

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerRouter.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    public class SignalRTriggerRouter
+    {
+        private readonly Dictionary<string, Dictionary<string, SignalRListener>> _listeners = new Dictionary<string, Dictionary<string, SignalRListener>>(StringComparer.OrdinalIgnoreCase);
```
realwanpengli: >_listeners [](start = 81, length = 10)

How about make it a class, so that we can easy know there's 2 levels dict, first hub level then method level.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBinding.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                _value = value;
+            }
+
+            public Task<object> GetValueAsync()
+            {
+                return Task.FromResult(_value);
+            }
+
+            public string ToInvokeString()
```
realwanpengli: >ToInvokeString [](start = 26, length = 14)

no reference?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBinding.cs
```diff
@@ -0,0 +1,114 @@
...
+        private class SignalRTriggerValueProvider : IValueBinder
+        {
+            private object _value;
+
+            public SignalRTriggerValueProvider(object value)
+            {
+                _value = value;
+            }
+
+            public Task<object> GetValueAsync()
```
realwanpengli: >GetValueAsync [](start = 32, length = 13)

no reference?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBinding.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+            
+            var functionNameAttribute = _parameterInfo.Member.GetCustomAttribute<FunctionNameAttribute>(false);
+            var methodName = functionNameAttribute.Name;
+
+            return Task.FromResult<IListener>(new SignalRListener(context.Executor, _router, _attribute.HubName, methodName));
```
realwanpengli: >SignalRListener [](start = 50, length = 15)

Looks like the key point is to add Excutor for a unique pair (hubname, methodname) into router, how about simple add Excutor to router and use an empty listner instead.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBinding.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>
+        /// Defined what other bindings can use and return value.
+        /// </summary>
+        private IReadOnlyDictionary<string, Type> CreateBindingContract()
+        {
+            var contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase)
+            {
+                // Functions can bind to parameter name "hubName" directly
+                { HubNameKey, typeof(string) },
```
realwanpengli: >HubNameKey [](start = 18, length = 10)

Invocation already have a property "HubName"

samples/simple-chat/csharp/FunctionApp/Functions.cs
```diff
@@ -59,6 +59,16 @@ public static class Functions
             await serviceHubContext.Clients.All.SendAsync("newMessage", message);
         }
 
+        // curl -X POST -H "Content-type: application/json" -d '{"HubName": "simplechat", "Data": {"Target": "broadcast2", "Arguments": ["hello"]}}' localhost:7071/runtime/webhooks/signalr/simplechat
+        [FunctionName("broadcast2")]
+        public static async Task BroadcastWithSignalRTrigger(
+            [SignalRTrigger(HubName = "simplechat")]InvocationContext context)
+        {
+            var message = (string)context.Data.Arguments[0];
+            var serviceHubContext = await StaticServiceHubContextStore.Get().GetAsync("simplechat");
```
realwanpengli: >simplechat [](start = 87, length = 10)

get it from InvocationContext?

realwanpengli: >StaticServiceHubContextStore [](start = 42, length = 28)

`StaticServiceHubContextStore` is only available for C#. 
For general languages, is it possible adding a new property in InvocationContext to tell what kind of methods (broadcast, echo, etc.) and send the message behind?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBinding.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
+        {
+            var bindingData = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+
+            if (value is SignalRTriggerEvent triggerEvent)
+            {
+                var bindingContext = triggerEvent.Context;
+                bindingData.Add(HubNameKey, bindingContext.HubName);
```
realwanpengli: >HubName [](start = 59, length = 7)

Invocation already have a property "HubName"

realwanpengli:
The function uses this trigger will not display information in console.
Can you add information like function using http trigger?

samples/simple-chat/csharp/FunctionApp/Functions.cs
davidfowl: Can we get rid of this API completely?

zackliu: @davidfowl  `InvocationContext` need to be available in various languages, it can only contain properties without methods. A solution is to build some SDKs in popular languages, especially `js` . Customers can just import these SDK packages instead of using output binding. Thus, they can use the similar pattern `..Clients.All.SendAsync(...)`. But it's another topic about how making output direction much easier to use. I think it's hard to integrate with the input Trigger.

PR #91, opened 2020-02-11
Add SignalR Trigger

zackliu:
It's a rough skeleton for signalr trigger. Seems it's way too large for a PR, I will see how to split it.


src/SignalRServiceExtension/TriggerBindings/Utils/JsonMessageParser.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    internal class JsonMessageParser : MessageParser
+    {
+        private const string TypePropertyName = "type";
+
+        public override bool TryParseMessage(ref ReadOnlySequence<byte> buffer, out ISignalRServerlessMessage message)
+        {
+            var jsonString = Encoding.UTF8.GetString(buffer.ToArray());
```
vwxyzh: using stream instead of create string.

PR #92, opened 2020-02-13
Add utils, context and constants for trigger

zackliu:
Add a separate project `Microsoft.Azure.SignalR.Serverless.Protocols`, which contains message context and protocols.

Current, I have to use MessagePack1.x, or we will depend one differenct messagepack version.
We have a plan to move the proj to SDK repo and then we can self-contained the messagepack2.x.


src/Microsoft.Azure.SignalR.Serverless.Protocols/Internal/MemoryBufferWriter.cs
```diff
@@ -0,0 +1,342 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR.Serverless.Protocols
+{
+    internal sealed class MemoryBufferWriter : Stream, IBufferWriter<byte>
```
vicancy: looks like this is copied from https://github.com/dotnet/aspnetcore/blob/master/src/SignalR/common/Shared/MemoryBufferWriter.cs, keep it as it is and add comment about the reference?

src/Microsoft.Azure.SignalR.Serverless.Protocols/JsonServerlessProtocol.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Buffers;
+using System.IO;
+using System.Text;
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.SignalR.Serverless.Protocols
+{
+    public class JsonServerlessProtocol : IServerlessProtocol
```
vicancy: this sounds more like a parser than a protocol?

zackliu: But `TryParseMessage` is also a functionality in SignalR protocol. It just miss the `GetMessageBytes` here.

src/Microsoft.Azure.SignalR.Serverless.Protocols/ServerlessMessage.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Microsoft.Azure.SignalR.Serverless.Protocols
+{
+    public abstract class ServerlessMessage
+    {
+        [JsonProperty(PropertyName = "type")]
+        public int Type { get; set; }
+    }
+
+    public class InvocationMessage : ServerlessMessage
```
JialinXin: >InvocationMessage [](start = 17, length = 17)

It seems in Runtime, `InvocationMessage` doesn't have type property, Shall we make RT and here consistent?

src/Microsoft.Azure.SignalR.Serverless.Protocols/Constants.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.SignalR.Serverless.Protocols
+{
+    internal static class ServerlessProtocolConstants
```
JialinXin: >ServerlessProtocolConstants [](start = 26, length = 27)

`PingMessage` is missed here. Not supported?

src/Microsoft.Azure.SignalR.Serverless.Protocols/ServerlessMessage.cs
zackliu: `InvocationMessage` in RT is not the `InvocationMessage` here. That in RT is just a wrapper over the real `InvocationMessage`

src/Microsoft.Azure.SignalR.Serverless.Protocols/Constants.cs
zackliu: No. Ping message is totally handled by RT.

src/Microsoft.Azure.SignalR.Serverless.Protocols/MessagePackServerlessProtocol.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Collections.Generic;
+using System.IO;
+
+using MessagePack;
+
+namespace Microsoft.Azure.SignalR.Serverless.Protocols
+{
+    public class MessagePackServerlessProtocol : IServerlessProtocol
+    {
+        public int Version => 1;
```
JialinXin: >Version [](start = 19, length = 7)

Just question, how will we use the Version? In case the parse method have an update in the future? Will this also supported in RT?

src/Microsoft.Azure.SignalR.Serverless.Protocols/MessagePackServerlessProtocol.cs
```diff
@@ -0,0 +1,52 @@
...
+                default:
+                    // TODO:OpenConnectionMessage and CloseConnectionMessage only will be sent in JSON format. It can be added later.
+                    message = null;
+                    break;
+            }
+
+            return message != null;
+        }
+
+        private static InvocationMessage ConvertInvocationMessage(byte[] input, ref int offset)
```
JialinXin: >ConvertInvocationMessage [](start = 41, length = 24)

Seems same as the method in MessagePackHelper. Can be merged?

zackliu: Oh, this should be removed in helper in `MessagePackHelper`.

src/Microsoft.Azure.SignalR.Serverless.Protocols/MessagePackServerlessProtocol.cs
zackliu: That's really a problem, we don't have a mechanism like handshake to negotiate a version like traditional signalr.

src/SignalRServiceExtension/TriggerBindings/Context/InvocationContext.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// The headers of request.
+        /// </summary>
+        public IDictionary<string, string> Headers { get; set; }
+
+        /// <summary>
+        /// The query of the request when client connect to the service.
+        /// </summary>
+        public IDictionary<string, string> Query { get; set; }
+
+
```
JialinXin: remove extra line

src/Microsoft.Azure.SignalR.Serverless.Protocols/IServerlessProtocol.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.SignalR.Serverless.Protocols
```
JialinXin: >Microsoft.Azure.SignalR.Serverless.Protocols [](start = 10, length = 44)

Another question, should this be part of Microsoft.Azure.SignalR.Protocols? Could it be used besides Azure Function? Will we publish a new package for this?

JialinXin: If internal only, shall we make the interfaces and class internal and make it visible internal?

---
In reply to: [380490961](https://github.com/Azure/azure-functions-signalrservice-extension/pull/92#discussion_r380490961) [](ancestors = 380490961)

src/Microsoft.Azure.SignalR.Serverless.Protocols/MessagePackServerlessProtocol.cs
JialinXin: Maybe add a TODO comment in the interface for future discussion?

---
In reply to: [380465249](https://github.com/Azure/azure-functions-signalrservice-extension/pull/92#discussion_r380465249) [](ancestors = 380465249)

src/Microsoft.Azure.SignalR.Serverless.Protocols/IServerlessProtocol.cs
zackliu: Yes, it need to be publish as a seperate package. However, I think it doesn't need to be a part of `Microsoft.Azure.SignalR.Protocols`. And we will move it to SDK repo later.

test/Microsoft.Azure.SignalR.Serverless.Protocols.Tests/Microsoft.Azure.SignalR.Serverless.Protocols.Tests.csproj
```diff
@@ -0,0 +1,18 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkPackageVersion)" />
+    <PackageReference Include="Moq" Version="$(MoqPackageVersion)" />
+    <PackageReference Include="xunit" Version="$(XunitPackageVersion)" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitRunnerVisualstudioPackageVersion)" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\SignalRServiceExtension\Microsoft.Azure.WebJobs.Extensions.SignalRService.csproj" />
```
JialinXin: >..\..\src\SignalRServiceExtension\Microsoft.Azure.WebJobs.Extensions.SignalRService.csproj [](start = 31, length = 90)

if the serverless.protocol project is planning to move out, shall we reduce the test denpendency here?

test/Microsoft.Azure.SignalR.Serverless.Protocols.Tests/MessagePackParseTests.cs
```diff
@@ -0,0 +1,61 @@
...
+            yield return new object[]
+            {
+                Guid.NewGuid().ToString(), Guid.NewGuid().ToString(),
+                new object[] {new object[] { null, Guid.NewGuid().ToString() }}
+            };
+        }
+
+        [Theory]
+        [MemberData(nameof(GetParameters))]
+        public void MessagePackParseTest(string invocationId, string target, object[] arguments)
```
JialinXin: >MessagePackParseTest [](start = 20, length = 20)

add test for OpenConnectionMessage and CloseConnectionMessage as well? And JsonParseTests?

zackliu: Added.

test/Microsoft.Azure.SignalR.Serverless.Protocols.Tests/Microsoft.Azure.SignalR.Serverless.Protocols.Tests.csproj
zackliu: Let's keep it as is now. We can have some clean up before moving

src/Microsoft.Azure.SignalR.Serverless.Protocols/IServerlessProtocol.cs
zackliu: It's not only for internal use. People who are not using Azure Function trigger may leverage it directly.

PR #94, opened 2020-02-27
Add Dispatcher and Executor for SignalR Trigger

zackliu:
Add dispatcher and executor to make the full workflow completed.
There are two flows:
* The binding flow (When webhost startup): Azure Function -> SignalRTriggerBinding.CreateListenerAsync() -> save Exectors in Dispatcher
* The triggering flow: SignalR Service -> Webhook Provider -> Dispacher -> SignalRMethodExecutor ->  SignalRTriggerBinding.BindAsync() -> Customer's Function


JialinXin:
> namespace Microsoft.Azure.WebJobs.Extensions.SignalRService.Exceptions

use `Microsoft.Azure.WebJobs.Extensions.SignalRService` directly?

---
Refers to: src/SignalRServiceExtension/Exceptions/FailedRouteEventException.cs:1 in de7326e. [](commit_id = de7326e4c9ea7f242e0dc7b243a3ab85046e2ea5, deletion_comment = False)

JialinXin:
> namespace Microsoft.Azure.WebJobs.Extensions.SignalRService.Exceptions

miss copyright.

---
Refers to: src/SignalRServiceExtension/Exceptions/FailedRouteEventException.cs:1 in de7326e. [](commit_id = de7326e4c9ea7f242e0dc7b243a3ab85046e2ea5, deletion_comment = False)

src/SignalRServiceExtension/TriggerBindings/Utils/TupleIgnoreCasesComparer.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    internal class TupleIgnoreCasesComparer: IEqualityComparer<(string, string, string)>
```
JialinXin: >Tuple [](start = 19, length = 5)

TupleStringIgnoreCaseComparer ?

src/SignalRServiceExtension/Exceptions/SignalRTriggerException.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
```
vicancy: copyright headers

src/SignalRServiceExtension/TriggerBindings/Resolver/DefaultRequestResolver.cs
```diff
@@ -0,0 +1,69 @@
...
+                !request.Headers.Contains(Constants.AsrsCategory) ||
+                !request.Headers.Contains(Constants.AsrsEvent) ||
+                !request.Headers.Contains(Constants.AsrsConnectionIdHeader))
+            {
+                context = null;
+                return false;
+            }
+
+            context = new InvocationContext();
+            context.ConnectionId = request.Headers.GetValues(Constants.AsrsConnectionIdHeader).First();
```
vicancy: use TryGetValue?

zackliu: If you think we should avoid multiple access to the same header value, I have a new apporach in the new commit. I think use GetValues().FirstOrDefault() is a much easier and cleaerer way than `TryGetValue`.

build/dependencies.props
```diff
@@ -14,6 +14,7 @@
     <MessagePackPackageVersion>1.9.11</MessagePackPackageVersion>
     <NewtonsoftJsonPackageVersion>11.0.2</NewtonsoftJsonPackageVersion>
     <SystemMemoryPackageVersion>4.5.3</SystemMemoryPackageVersion>
+    <MicrosoftAspNetCoreMvcWebApiCompatShimPackageVersion>2.2.0</MicrosoftAspNetCoreMvcWebApiCompatShimPackageVersion>
```
davidfowl: This should not be in the list of dependencies. Why do we need it. It's deprecated in 3.0...

src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
```diff
@@ -4,19 +4,23 @@
...
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
-    [Extension("SignalR")]
-    internal class SignalRConfigProvider : IExtensionConfigProvider
+    [Extension("SignalR", "signalr")]
+    internal class SignalRConfigProvider : IExtensionConfigProvider, IAsyncConverter<HttpRequestMessage, HttpResponseMessage>
```
davidfowl: Why are we using HttpRequestMessage and HttpResponseMessage for everything?

zackliu: It's Azure Function webhook provider's requirement. It pass HttpRequestMessage to us and receive HttpResponseMessage as return.

build/dependencies.props
zackliu: I didn't notice that. Now it's just used in tests. I use `new HttpRequestMessageFeature().HttpRequestMessage` to get HttpRequestMessage in tests.

src/SignalRServiceExtension/TriggerBindings/Resolver/DefaultRequestResolver.cs
```diff
@@ -0,0 +1,69 @@
...
+            context = new InvocationContext();
+            // Required properties
+            context.ConnectionId = request.Headers.GetValues(Constants.AsrsConnectionIdHeader).FirstOrDefault();
+            context.Hub = request.Headers.GetValues(Constants.AsrsHubNameHeader).FirstOrDefault();
+            context.Category = request.Headers.GetValues(Constants.AsrsCategory).FirstOrDefault();
+            context.Event = request.Headers.GetValues(Constants.AsrsEvent).FirstOrDefault();
+            if (string.IsNullOrEmpty(context.ConnectionId) ||
+                string.IsNullOrEmpty(context.Hub) ||
+                string.IsNullOrEmpty(context.Category) ||
+                string.IsNullOrEmpty(context.Event))
```
JialinXin: Some checks already covered in `SignalRTriggerDispater.ExecuteAsync()`, could be reduced here?

build/dependencies.props
zackliu: Microsoft.AspNetCore.Mvc.WebApiCompatShim seems still be used in Azure Function, even though v3.0. https://github.com/Azure/azure-functions-host/blob/v3.0.13152/src/WebJobs.Script.WebHost/Controllers/HostController.cs#L14

src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
wanlwanl: Can we also support HttpRequest? I see HttpTrigger support both

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerAttribute.cs
```diff
@@ -11,6 +11,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
     [Binding]
     public class SignalRTriggerAttribute : Attribute
     {
+        [AppSetting(Default = Constants.AzureSignalRConnectionStringName)]
```
wanlwanl: add summary

src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
zackliu: @wanlwanl  As far as I know, `HttpTrigger` is so different that it highly integrate with Azure Function. So, `HttpTrigger` has the "privilege" over other triggers. `HttpTrigger` can design the route path, but if we use webhookProvider, we need keep there convension and the url path must be "host/runtime/webhook/signalr".

test/SignalRServiceExtension.Tests/Trigger/SignalRMethodExecutorTests.cs
```diff
@@ -0,0 +1,129 @@
...
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Moq;
+using Newtonsoft.Json;
+using SignalRServiceExtension.Tests.Utils;
+using Xunit;
+using ExecutionContext = Microsoft.Azure.WebJobs.Extensions.SignalRService.ExecutionContext;
+
+namespace SignalRServiceExtension.Tests.Trigger
+{
+    public class SignalRMethodExecutorTests
```
JialinXin: >SignalRMethodExecutorTests [](start = 17, length = 26)

Add some tests to cover IRequestResolver return validation false as well?

src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
```diff
@@ -4,19 +4,23 @@
...
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
-    [Extension("SignalR")]
-    internal class SignalRConfigProvider : IExtensionConfigProvider
+    [Extension("SignalR", "signalr")]
```
JialinXin: >signalr [](start = 27, length = 7)

Is this case sensitive? Is there better compatible ways instead of list all? how about `signalR` which also might be a usual naming convention way?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerDispatcher.cs
```diff
@@ -0,0 +1,98 @@
...
+                return new HttpResponseMessage(HttpStatusCode.BadRequest);
+            }
+            
+            if (_executors.TryGetValue(key, out var executor))
+            {
+                try
+                {
+                    return await executor.ExecuteAsync(req);
+                }
+                catch (SignalRTriggerException ex)
```
JialinXin: >SignalRTriggerException [](start = 23, length = 23)

Could some general exception throws from `ITriggeredFunctionExecutor`? Shall we handle as well?

src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
zackliu: The second parameter is used in Url path(runtime/webhook/signalr), so I think lower case makes sense.

JialinXin: Get it. So in RT does the url accepts lowercase only? Could it somehow be modified to not normalized ones and be blocked here?

---
In reply to: [385643266](https://github.com/Azure/azure-functions-signalrservice-extension/pull/94#discussion_r385643266) [](ancestors = 385643266)

zackliu: No, RT can accept any case because customer may not use Azure Function. But for the trigger, from customer's view, the url is generate by trigger and ask them to copy/paste. I think it's OK.

JialinXin: Good.

---
In reply to: [385649396](https://github.com/Azure/azure-functions-signalrservice-extension/pull/94#discussion_r385649396) [](ancestors = 385649396)

src/SignalRServiceExtension/TriggerBindings/Executor/SignalRMethodExecutor.cs
```diff
@@ -0,0 +1,66 @@
...
+            if (!Resolver.ValidateSignature(request, executor.AccessKey))
+            {
+                //TODO: More detailed exception
+                throw new SignalRTriggerException();
+            }
+
+            return ExecuteAsync(executor.Executor, context, tcs);
+        }
+
+        private async Task<FunctionResult> ExecuteAsync(ITriggeredFunctionExecutor executor, InvocationContext context, TaskCompletionSource<object> tcs)
```
JialinXin: >ExecuteAsync [](start = 43, length = 12)

How about rename to `ExcuteAsyncCore`, kind of confusing with the abtract method `ExecuteAsync`.

src/SignalRServiceExtension/TriggerBindings/Resolver/DefaultRequestResolver.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+using Microsoft.AspNetCore.SignalR.Protocol;
+using Microsoft.Azure.SignalR.Serverless.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    internal class DefaultRequestResolver : IRequestResolver
```
JialinXin: >DefaultRequestResolver [](start = 19, length = 22)

Will IRequestResolver be public? If not, probably rename this class to SignalRRequstResolver is better.

test/SignalRServiceExtension.Tests/Trigger/SignalRMethodExecutorTests.cs
zackliu: I added some, and put it in the DispatcherTest

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerDispatcher.cs
zackliu: Updated. We will handle different details exceptions and response different result

src/SignalRServiceExtension/TriggerBindings/Executor/SignalRMethodExecutor.cs
zackliu: Sure

src/SignalRServiceExtension/TriggerBindings/Resolver/DefaultRequestResolver.cs
zackliu: Updated

PR #98, opened 2020-03-06
Add signature validation support

zackliu:
Add signature validation and meanwhile I found some provided function or attribute are actually not working. So, some updates about resolving property and validation are moved.

### DIFF
--- a/src/SignalRServiceExtension/TriggerBindings/Resolver/SignalRRequestResolver.cs
+++ b/src/SignalRServiceExtension/TriggerBindings/Resolver/SignalRRequestResolver.cs
@@ -5,6 +5,8 @@ using System;
 using System.Buffers;
 using System.Linq;
 using System.Net.Http;
+using System.Security.Cryptography;
+using System.Text;
 using System.Threading.Tasks;
 
 using Microsoft.AspNetCore.SignalR.Protocol;
@@ -26,8 +28,23 @@ namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 
         public bool ValidateSignature(HttpRequestMessage request, string accessToken)
         {
-            //TODO: Add real signature validation
-            return true;
+            if (!string.IsNullOrEmpty(accessToken) &&
+                request.Headers.TryGetValues(Constants.AsrsSignature, out var values))
+            {
+                var signatures = SignalRTriggerUtils.GetSignatureList(values.FirstOrDefault());
+                if (signatures == null)
+                {
+                    return false;
+                }
+                using (var hmac = new HMACSHA256(Encoding.UTF8.GetBytes(accessToken)))
+                {
+                    var hashBytes = hmac.ComputeHash(Encoding.UTF8.GetBytes(request.Headers.GetValues(Constants.AsrsConnectionIdHeader).First()));
+                    var hash = "sha256=" + BitConverter.ToString(hashBytes).Replace("-", "");
+                    return signatures.Contains(hash, StringComparer.OrdinalIgnoreCase);
+                }
+            }
+
+            return false;
         }
 
         public bool TryGetInvocationContext(HttpRequestMessage request, out InvocationContext context)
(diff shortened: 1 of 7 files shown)


src/SignalRServiceExtension/TriggerBindings/Resolver/SignalRRequestResolver.cs
```diff
@@ -26,8 +28,23 @@ public bool ValidateContentType(HttpRequestMessage request)
...
+            if (!string.IsNullOrEmpty(accessToken) &&
+                request.Headers.TryGetValues(Constants.AsrsSignature, out var values))
+            {
+                var signatures = SignalRTriggerUtils.GetSignatureList(values.FirstOrDefault());
+                if (signatures == null)
+                {
+                    return false;
+                }
+                using (var hmac = new HMACSHA256(Encoding.UTF8.GetBytes(accessToken)))
+                {
```
JialinXin: Add comments to aware this algorithm should be synced with RT.

JialinXin:
>             var attribute = parameterInfo.GetCustomAttribute<SignalRTriggerAttribute>(false);

why here explicit assign false inherit but inside `GetResolvedConnectionString` doesn't?

---
Refers to: src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs:37 in a331253. [](commit_id = a3312533cd75c6cfebb7f8ad381fd1c129b339f8, deletion_comment = False)

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -33,8 +39,68 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
+                }
+                resolvedConnectionString = _nameResolver.Resolve(attribute.Default);
+            }
+
+            if (string.IsNullOrEmpty(resolvedConnectionString))
+            {
+                resolvedConnectionString = _options.ConnectionString;
+            }
+
+            return resolvedConnectionString;
```
JialinXin: return string.IsNullOrEmpty(resolvedConnectionString) ? _options.ConnectionString : resolvedConnectionString;

wanlwanl: use the Utils.FirstOrDefault?

src/SignalRServiceExtension/TriggerBindings/Resolver/SignalRRequestResolver.cs
```diff
@@ -24,10 +35,31 @@ public bool ValidateContentType(HttpRequestMessage request)
...
+
+            if (!string.IsNullOrEmpty(accessToken) &&
+                request.Headers.TryGetValues(Constants.AsrsSignature, out var values))
+            {
+                var signatures = SignalRTriggerUtils.GetSignatureList(values.FirstOrDefault());
+                if (signatures == null)
+                {
+                    return false;
+                }
+                using (var hmac = new HMACSHA256(Encoding.UTF8.GetBytes(accessToken)))
```
vicancy: maybe move to the protocol project which will be shared by both side?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
zackliu: But there're only two parameters, I think it's simple and efficient enough to use `?:`

src/SignalRServiceExtension/TriggerBindings/Resolver/SignalRRequestResolver.cs
zackliu: As it's a common HMAC algorithm, and there're existing implementation for each language. I don't put it in protocol for now.

PR #100, opened 2020-03-09
Fix some issues

zackliu:
Fix some minor issues

### DIFF
--- a/src/SignalRServiceExtension/TriggerBindings/Utils/SignalRTriggerUtils.cs
+++ b/src/SignalRServiceExtension/TriggerBindings/Utils/SignalRTriggerUtils.cs
@@ -55,7 +55,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 
             // The query string looks like "?key1=value1&key2=value2"
             var queryArray = queryString.TrimStart('?').Split(QuerySeparator, StringSplitOptions.RemoveEmptyEntries);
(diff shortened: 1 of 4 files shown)


src/SignalRServiceExtension/TriggerBindings/Utils/SignalRTriggerUtils.cs
```diff
@@ -65,9 +66,9 @@ public static string GetAccessKey(string connectionString)
                 return default;
             }
 
-            // The claim string looks like "a= v, b= v"
-            return claims.Split(HeaderSeparator)
-                .Select(p => p.Split(ClaimsSeparator, StringSplitOptions.RemoveEmptyEntries))
+            // The claim string looks like "a: v, b: v"
+            return claims.Split(HeaderSeparator, StringSplitOptions.RemoveEmptyEntries)
+                .Select(p => p.Split(ClaimsSeparator, StringSplitOptions.RemoveEmptyEntries)).Where(l => l.Length == 2)
```
JialinXin: >ClaimsSeparator [](start = 37, length = 15)

what if value contains `:`? will it destroy the structure?

zackliu: I use `: ` as a separator. But if there a `: ` in claim, it will be destroyed. Do we have other options?

PR #101, opened 2020-03-10
Add a bidirectional chat sample to demonstrate SignalR Trigger

zackliu:
Here is the C# chat sample to demonstrate new feature: SignalR Trigger
The trigger can collaborate with SignalR Service's new upstream feature smoothly to empower customer to handle messages from service in serverless mode.

Now, client can call `connection.send()` or `connection.invoke()` just like what they usually do in default mode. 
Also, in C#, Function can design there own attribute to implement Authorization. In the sample, a very simple attribute that only allow the admin to call broadcast is demonstrated.

It's not the final design, please give some suggestions. And I will add a JavaScript sample later.


samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -0,0 +1,170 @@
...
+            await signalRMessages.AddAsync(new SignalRMessage()
+            {
+                Target = NewConnectionTarget,
+                Arguments = new object[] { new NewConnection(invocationContext.ConnectionId) }
+            });
+        }
+
+        [FunctionAuthorize]
+        [FunctionName("broadcast")]
+        public static async Task Broadcast([SignalRTrigger(Hub, Category.Messages, "broadcast", "message")]InvocationContext invocationContext,
```
JialinXin: >broadcast [](start = 84, length = 9)

How about enum messages event as well. the functionality is limited to our supported APIs.

samples/bidirectional-chat/csharp/local.settings.sample.json
```diff
@@ -0,0 +1,15 @@
+{
+  "IsEncrypted": false,
+  "Values": {
+    "AzureWebJobsStorage": "",
+    "AzureWebJobsDashboard": "",
+    "FUNCTIONS_WORKER_RUNTIME": "dotnet",
+    "AzureSignalRConnectionString": "Endpoint=http://localhost;AccessKey=ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789ABCDEFGH;Version=1.0;Port=8080",
```
JialinXin: hide value?

samples/bidirectional-chat/csharp/Authorize/FunctionAuthorizeAttribute.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace bidirectional_chat
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+#pragma warning disable CS0618 // Type or member is obsolete
+    internal class FunctionAuthorizeAttribute: FunctionInvocationFilterAttribute
+    {
+        private const string AdminKey = "admin";
+
+        public override Task OnExecutingAsync(FunctionExecutingContext executingContext, CancellationToken cancellationToken)
```
chenkennt: Can we extract some common logic into a base class so that developer only needs to focus on the claim validation part.

samples/bidirectional-chat/csharp/Function.cs
chenkennt: Is it possible to get method name from FunctionName attribute?

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -0,0 +1,170 @@
...
+        public static SignalRConnectionInfo GetSignalRInfo(
+            [HttpTrigger(AuthorizationLevel.Anonymous)] HttpRequest req,
+            [SignalRConnectionInfo(HubName = Hub, UserId = "{headers.x-ms-signalr-user-id}")] SignalRConnectionInfo connectionInfo)
+        {
+            return connectionInfo;
+        }
+
+        [FunctionName("connect")]
+        public static async Task Connect([SignalRTrigger(Hub, Category.Connections, Event.Connect)]InvocationContext invocationContext,
+            [SignalR(HubName = Hub)]IAsyncCollector<SignalRMessage> signalRMessages)
```
chenkennt: Both SignalRTrigger and SignalR output has hub name. Is it possible to remove one of them?

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -0,0 +1,170 @@
...
+            await signalRMessages.AddAsync(new SignalRMessage()
+            {
+                UserId = userName,
+                Target = NewMessageTarget,
+                Arguments = new[] { new NewMessage(invocationContext, message) }
+            });
+        }
+
+        [FunctionName("sendToConnection")]
+        public static async Task SendToConnection([SignalRTrigger(Hub, Category.Messages, "sendToConnection", "connectionId", "message")]InvocationContext invocationContext,
```
chenkennt: How about move parameter name to be in front of each parameter. Something like this:

```cs
Task SendToConnection(..., [SignalRParameter("message")]message, [SignalRParameter("connectionId")]connectionId)
```

davidfowl:
The way I see it, we need to get as close to possible to a Hub as we can on with the Azure functions infrastructure and see what that looks like in multiple languages. I have some ideas of what a nice node and C# experience looks like.

samples/bidirectional-chat/csharp/Function.cs
zackliu: The function name need to be globally unique, but method name just need to be unique in hub level

zackliu:
@davidfowl I've just updated the sample and applied class base on it.

Briefly, when Functions are defined in under a class that derived from `ServerlessHub`, many configurations work can be simplified.
```csharp
[FunctionName(nameof(SendToConnection))]
public async Task SendToConnection([SignalRTrigger]InvocationContext invocationContext,
    [SignalRParameter]string connectionId,
    [SignalRParameter]string message)
{
    await Clients.Client(connectionId).SendAsync(NewMessageTarget, new NewMessage(invocationContext, message));
}
```
Considered the balance of convenience and correction, `[SignalRParameter]` is used to make it more controllable and avoid conflict with other binding's binding data.

However, all these simplify can only be applied in C#. As the limitation of Function architecture, we can do less in other language. I will then open another PR to demonstrate trigger in nodejs.

src/SignalRServiceExtension/TriggerBindings/ServerlessHub.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// When a class derived from <see cref="ServerlessHub"/>,
+    /// all the method in the are identified as using class based model.
+    /// </summary>
+    public abstract class ServerlessHub : IDisposable
+    {
+        private bool _disposed;
+
+        public ServerlessHub()
+        {
+            var hubName = GetType().Name;
```
JialinXin: >hubName [](start = 16, length = 7)

add HubName to public immutable property? So that in functions, negotiation method could get the value as well?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -39,20 +38,54 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
+            // We have to model for C#, one is function based model which also work in multiple language
+            // Another one is class based model, which is highly close to SignalR itself but must keep some conventions.
+            var method = (MethodInfo) parameterInfo.Member;
+            var declaredType = method.DeclaringType;
+            // Class based model
+            if (declaredType != null && declaredType.IsSubclassOf(typeof(ServerlessHub)))
+            {
+                hubName = string.IsNullOrEmpty(hubName) ? declaredType.Name : hubName;
+                category = string.IsNullOrEmpty(category) ? GetCategoryFromMethodName(method.Name) : category;
+                @event = string.IsNullOrEmpty(@event) ? method.Name : @event;
```
JialinXin: is here we can already detect if both attribute and parameter used and throw exception early, maybe return for non class model as well to save some process below?

src/SignalRServiceExtension/TriggerBindings/ServerlessHub.cs
zackliu: They are in the different pipelines. The class instance constructs every time the function is invoked. So, the properties of the class cannot be used in attribute parameters. If the negotiation method need to get hub name, it also needs to use `MethodInfo` in binding.

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -0,0 +1,124 @@
...
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.WebJobs.Extensions.SignalRService;
+using Microsoft.AspNetCore.SignalR;
+
+namespace bidirectional_chat
+{
+    public class SimpleChat : ServerlessHub
+    {
+        private const string Hub = "simplechat";
```
JialinXin: >"simplechat" [](start = 35, length = 12)

use GetType().Name to uniform the method of getting hub value?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -39,20 +38,54 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
             }
-            var resolvedAttribute = GetParameterResolvedAttribute(attribute);
+            var resolvedAttribute = GetParameterResolvedAttribute(attribute, parameterInfo);
             ValidateSignalRTriggerAttributeBinding(resolvedAttribute);
             
             return Task.FromResult<ITriggerBinding>(new SignalRTriggerBinding(parameterInfo, resolvedAttribute, _dispatcher));
         }
 
-        private SignalRTriggerAttribute GetParameterResolvedAttribute(SignalRTriggerAttribute attribute)
+        private SignalRTriggerAttribute GetParameterResolvedAttribute(SignalRTriggerAttribute attribute, ParameterInfo parameterInfo)
```
JialinXin: >GetParameterResolvedAttribute [](start = 40, length = 29)

add UTs for this change?

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -0,0 +1,124 @@
...
+        private const string NewMessageTarget = "newMessage";
+        private const string NewConnectionTarget = "newConnection";
+
+        [FunctionName("negotiate")]
+        public static SignalRConnectionInfo GetSignalRInfo(
+            [HttpTrigger(AuthorizationLevel.Anonymous)] HttpRequest req,
+            [SignalRConnectionInfo(HubName = Hub, UserId = "{headers.x-ms-signalr-user-id}")] SignalRConnectionInfo connectionInfo)
+        {
+            return connectionInfo;
```
wanlwanl: looks like the "admin" claim doesn't send to RT

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        [FunctionName(nameof(Connect))]
+        public async Task Connect([SignalRTrigger]InvocationContext invocationContext)
```
wanlwanl: Can we change the name Connect/Disconnect to OnConnected/OnDisconnected? The server SDK uses onConnectedAsync/onDisconnectedAsync

zackliu: But the `connect` is event is triggered before the client is really connected. It's in the handshake step which is different from SDK.

samples/bidirectional-chat/csharp/Function.cs
zackliu: Yes, I would like to add the claim by using `IdToken` and `ClaimTypeList` but now the binding expression doesn't work now.

samples/bidirectional-chat/csharp/Function.cs
wanlwanl: When we verify the connection whether it comes from ASRS, if it is then pass, if not then drop, seems don't need customer to handle. 

`Connect` looks confused to CX. Can we handle it internally? If we want to support custom auth, we can use the DI feature (in custom auth PR). 

How about add a `Connected` event when the connection is really connected? and if we want to drop the connection, add an `Abort` event, which is similar to server SDK?

samples/bidirectional-chat/csharp/Authorize/FunctionAuthorizeAttribute.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.Azure.WebJobs.Extensions.SignalRService;
+using Microsoft.Azure.WebJobs.Host;
+
+namespace bidirectional_chat
```
wanlwanl: namespace looks different from others...

samples/bidirectional-chat/csharp/Function.cs
davidfowl: Yes!

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
zackliu: Added

samples/bidirectional-chat/csharp/Function.cs
zackliu: Updated. Now `connect` and `disconnect` are renamed to `connected` and `disconnected`. Also, when `connected` invoked, the client has already connected.
May be there need to have a separate event `auth` or `handshake`, but it's not necessary now.

davidfowl:
Can we get rid of [SignalRParameter] and move InvocationContext to a Property on the base class?

zackliu:
@davidfowl Getting rid of `[SignalRParameter]` makes the trigger hand to decide which parameter the customer want to bind by signar tigger. Customers usually use other input/output bindings in Azure Function. Using `[SignalRParameter]` is a little tedious but clear and easy for learning.
`InvocationContext` is directly passed to the Function method. Seems it's hard to get it in the ctor and assign to a Property.

samples/bidirectional-chat/csharp/Authorize/FunctionAuthorizeAttribute.cs
zackliu: That's good. I will use another PR to do that.

davidfowl:
They want to bing everything that isn’t already marked. Mark the things that are non signalr explicitly.

src/SignalRServiceExtension/TriggerBindings/ServerlessHub.cs
```diff
@@ -0,0 +1,54 @@
...
+            var hubContext = StaticServiceHubContextStore.Get().GetAsync(hubName).GetAwaiter().GetResult();
+            Clients = hubContext.Clients;
+            Groups = hubContext.Groups;
+            UserGroups = hubContext.UserGroups;
+        }
+
+        public IHubClients Clients { get; set; }
```
chenkennt: private set?

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -0,0 +1,124 @@
...
+namespace FunctionApp
+{
+    public class SimpleChat : ServerlessHub
+    {
+        private const string Hub = nameof(SimpleChat);
+        private const string NewMessageTarget = "newMessage";
+        private const string NewConnectionTarget = "newConnection";
+
+        [FunctionName("negotiate")]
+        public static SignalRConnectionInfo GetSignalRInfo(
```
chenkennt: Can this be moved into ServerlessHub?

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -0,0 +1,124 @@
...
+        [FunctionName(nameof(LeaveUserFromGroup))]
+        public async Task LeaveUserFromGroup([SignalRTrigger]InvocationContext invocationContext,
+            [SignalRParameter]string userName,
+            [SignalRParameter]string groupName)
+        {
+            await UserGroups.RemoveFromGroupAsync(userName, groupName);
+        }
+
+        [FunctionName(nameof(Disconnect))]
+        public void Disconnect([SignalRTrigger]InvocationContext invocationContext)
```
davidfowl: Can you call these methods explicitly?

zackliu: What's your "explicitly" means?

src/SignalRServiceExtension/TriggerBindings/SignalRParameterAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Mark the parameter as the SignalR parameter that need to bind arguments.
+    /// It's mutually exclusive with <see cref="SignalRTriggerAttribute.ParameterNames"/>. That means
+    /// you can not set <see cref="SignalRTriggerAttribute.ParameterNames"/> and use <see cref="SignalRParameterAttribute"/>
+    /// at the same time.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public class SignalRParameterAttribute : Attribute
+    {
```
chenkennt: Also allow specify parameter name

zackliu: I think it's useless. The parameter name is used to match the actual parameter name in method. It should be exactly the name it decorates. We should avoid user using like
```csharp
void Function([SignalRTrigger]InvocationContext context, [SignalRParameter("arg1")]arg0, [SignalRParameter("arg0")]arg1)
```
It's just cause confusing.

samples/bidirectional-chat/csharp/Function.cs
davidfowl: Can these calls come from outside the signalr service?

davidfowl: Can the client trigger a call to disconnect and not be disconnected?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -39,20 +38,54 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
+                parameterNamesFromAttribute.Add(parameter.Name);
+            }
+
+            if (parameterNamesFromAttribute.Count != 0 && parameterNames.Length != 0)
+            {
+                throw new InvalidOperationException(
+                    $"{nameof(SignalRTriggerAttribute)}.{nameof(SignalRTriggerAttribute.ParameterNames)} and {nameof(SignalRParameterAttribute)} can not be set in the same Function.");
+            }
+
+            parameterNames = parameterNamesFromAttribute.Count != 0
```
chenkennt: Should check `parameterNames` first? `parametersNames` should have higher priority than attributes.

samples/bidirectional-chat/csharp/Function.cs
zackliu: Yes, as long as the message sent to SignalR trigger can follow our serverless spec.

samples/bidirectional-chat/content/index.html
```diff
@@ -0,0 +1,239 @@
...
+        };
+        return config;
+      }
+      let counter = 0;
+      function onNewMessage(message) {
+        message.id = counter++; // vue transitions need an id
+        data.messages.unshift(message);
+      };
+      function onNewConnection(message) {
+        data.myConnectionId = message.ConnectionId;
```
chenkennt: Do we want to do a camel case conversion for class members? (ConnectionId -> connectionId) I think SignalR by default does this conversion.

samples/bidirectional-chat/csharp/Function.cs
zackliu: If using SignalR Service, it won't happen. Because all message from client's invocation will be in the `messages` category. And connected/disconnected are `connections` category

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
zackliu: I don't want `parametersNames`  and attribute are set together. So, there's no priority issue.

samples/bidirectional-chat/csharp/Function.cs
zackliu: That's good. But it needs some investigation. I will put it into another RP.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -99,5 +146,34 @@ private void ValidateParameterNames(string[] parameterNames)
                 throw new ArgumentException("Elements in ParameterNames should be ignore case unique.");
             }
         }
+
+        private bool IsLegalClassBasedParameter(ParameterInfo parameter)
+        {
+            // In class based model, we treat all the parameters as a legal parameter exception the cases below
```
JialinXin: >exception [](start = 86, length = 9)

typo: except

src/SignalRServiceExtension/TriggerBindings/SignalRIgnoreAttribute.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    /// <summary>
+    /// In class based model, mark the parameter explicitly not to be a SignalR parameter.
+    /// That means it won't be bind to a InvocationMessage argument.
```
JialinXin: >be bind [](start = 28, length = 7)

typo: bind or be bound

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        [FunctionName(nameof(LeaveUserFromGroup))]
+        public async Task LeaveUserFromGroup([SignalRTrigger]InvocationContext invocationContext, string userName, string groupName)
+        {
+            await UserGroups.RemoveFromGroupAsync(userName, groupName);
+        }
+
+        [FunctionName(nameof(Disconnect))]
+        public void Disconnect([SignalRTrigger]InvocationContext invocationContext)
```
chenkennt: Should be `disconnected`?

zackliu: Use another PR to do that change including `onconnected`

src/SignalRServiceExtension/TriggerBindings/SignalRParameterAttribute.cs
chenkennt: Like in JSON serialization, you can use JsonProperty to give property a different name (for example a short name). But seems SignalR itself doesn't support this as well. So it should be ok to not support it for now.

chenkennt: Then do we really need this? A parameter is by default recognized as a SignalR parameter.

zackliu: Only in class base. It prepared for non-class based function. It's still better than using `ParameterNames` directly

src/SignalRServiceExtension/TriggerBindings/ServerlessHub.cs
```diff
@@ -0,0 +1,54 @@
...
+        public ServerlessHub()
+        {
+            var hubName = GetType().Name;
+            var hubContext = StaticServiceHubContextStore.Get().GetAsync(hubName).GetAwaiter().GetResult();
+            Clients = hubContext.Clients;
+            Groups = hubContext.Groups;
+            UserGroups = hubContext.UserGroups;
+        }
+
+        public IHubClients Clients { get; }
```
chenkennt: How about also add them to InvocationContext?

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -0,0 +1,112 @@
...
+        [FunctionName("negotiate")]
+        public static SignalRConnectionInfo GetSignalRInfo(
+            [HttpTrigger(AuthorizationLevel.Anonymous)] HttpRequest req,
+            [SignalRConnectionInfo(HubName = Hub, UserId = "{headers.x-ms-signalr-user-id}")] SignalRConnectionInfo connectionInfo)
+        {
+            return connectionInfo;
+        }
+
+        [FunctionName(nameof(Connected))]
+        public async Task Connected([SignalRTrigger]InvocationContext invocationContext, ILogger logger)
```
chenkennt: Can you still specify hub name in SignalRTrigger?

PR #103, opened 2020-03-12
Add a JavaScript sample to demonstrate SignalR Trigger

zackliu:
Related to #101 

The RP demonstrate SignalR Trigger usage in JavaScript. As far as I know, non-C# Functions has more limitation, so that binding data can not be bond to function's parameter directly. Instead, binding data are retrieved through `context.bindingData.<parameter-name>`.


zackliu:
We can develop a serverless SDK for node just like what we've done for C#. Then we can git rid of output binding and use syntax like `chat.clients.all.send('send', message);`.
However, for the input side, unless there's an entrypoint for us to customize cross-language invocation, the experience hard to be improved.

JialinXin (approved):

PR #108, opened 2020-03-17
Improvements in the class based model

zackliu:
Some improvements to make class based model much closer to SignalR SDK. Also make it more convenient.
* Add a more convenient negotiate approach which can assign claims in runtime.
* Use `OnConnected` and `OnDisconnected` to match that in SDK.
* Extract a `SignalRFilterAttribute` for easier way to implement a customer filter.

Also, design a much strict convention that class based model must use non-arg constructor.

### DIFF
--- a/samples/bidirectional-chat/csharp/Function.cs
+++ b/samples/bidirectional-chat/csharp/Function.cs
@@ -12,25 +12,23 @@ namespace FunctionApp
 {
     public class SimpleChat : ServerlessHub
     {
-        private const string Hub = nameof(SimpleChat);
         private const string NewMessageTarget = "newMessage";
         private const string NewConnectionTarget = "newConnection";
 
         [FunctionName("negotiate")]
-        public static SignalRConnectionInfo GetSignalRInfo(
-            [HttpTrigger(AuthorizationLevel.Anonymous)] HttpRequest req,
-            [SignalRConnectionInfo(HubName = Hub, UserId = "{headers.x-ms-signalr-user-id}")] SignalRConnectionInfo connectionInfo)
+        public SignalRConnectionInfo Negotiate([HttpTrigger(AuthorizationLevel.Anonymous)]HttpRequest req)
         {
-            return connectionInfo;
+            return Negotiate(req.Headers["x-ms-signalr-user-id"], GetClaims(req.Headers["Authorization"]));
         }
 
-        [FunctionName(nameof(Connected))]
-        public async Task Connected([SignalRTrigger]InvocationContext invocationContext, ILogger logger)
+        [FunctionName(nameof(OnConnected))]
+        public async Task OnConnected([SignalRTrigger]InvocationContext invocationContext, ILogger logger)
         {
             await Clients.All.SendAsync(NewConnectionTarget, new NewConnection(invocationContext.ConnectionId));
             logger.LogInformation($"{invocationContext.ConnectionId} has connected");
         }
 
+        [FunctionAuthorize]
         [FunctionName(nameof(Broadcast))]
         public async Task Broadcast([SignalRTrigger]InvocationContext invocationContext, string message, ILogger logger)
         {
@@ -80,8 +78,8 @@ namespace FunctionApp
             await UserGroups.RemoveFromGroupAsync(userName, groupName);
         }
 
-        [FunctionName(nameof(Disconnect))]
-        public void Disconnect([SignalRTrigger]InvocationContext invocationContext)
+        [FunctionName(nameof(OnDisconnect))]
+        public void OnDisconnect([SignalRTrigger]InvocationContext invocationContext)
         {
         }
 
--- a/samples/bidirectional-chat/csharp/Function.cs
+++ b/samples/bidirectional-chat/csharp/Function.cs
@@ -78,8 +78,8 @@ namespace FunctionApp
             await UserGroups.RemoveFromGroupAsync(userName, groupName);
         }
 
-        [FunctionName(nameof(OnDisconnect))]
-        public void OnDisconnect([SignalRTrigger]InvocationContext invocationContext)
+        [FunctionName(nameof(OnDisconnected))]
+        public void OnDisconnected([SignalRTrigger]InvocationContext invocationContext)
         {
         }
 
--- a/src/SignalRServiceExtension/TriggerBindings/InvocationContextExtensions.cs
+++ b/src/SignalRServiceExtension/TriggerBindings/InvocationContextExtensions.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Azure.SignalR.Management;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    public static class InvocationContextExtensions
+    {
+        public static async Task<IHubClients> GetClientsAsync(this InvocationContext invocationContext)
+        {
+            return (await StaticServiceHubContextStore.Get().GetAsync(invocationContext.Hub)).Clients;
+        }
+
+        public static async Task<IGroupManager> GetGroupsAsync(this InvocationContext invocationContext)
+        {
+            return (await StaticServiceHubContextStore.Get().GetAsync(invocationContext.Hub)).Groups;
+        }
+
+        public static async Task<IUserGroupManager> GetUserGroupManagerAsync(this InvocationContext invocationContext)
+        {
+            return (await StaticServiceHubContextStore.Get().GetAsync(invocationContext.Hub)).UserGroups;
+        }
+    }
+}
--- a/src/SignalRServiceExtension/TriggerBindings/InvocationContextExtensions.cs
+++ b/src/SignalRServiceExtension/TriggerBindings/InvocationContextExtensions.cs
@@ -9,16 +9,25 @@ namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
     public static class InvocationContextExtensions
     {
+        /// <summary>
+        /// Gets an object that can be used to invoke methods on the clients connected to this hub.
+        /// </summary>
         public static async Task<IHubClients> GetClientsAsync(this InvocationContext invocationContext)
         {
             return (await StaticServiceHubContextStore.Get().GetAsync(invocationContext.Hub)).Clients;
         }
 
+        /// <summary>
+        /// Get the group manager of this hub.
+        /// </summary>
         public static async Task<IGroupManager> GetGroupsAsync(this InvocationContext invocationContext)
         {
             return (await StaticServiceHubContextStore.Get().GetAsync(invocationContext.Hub)).Groups;
         }
 
+        /// <summary>
+        /// Get the user group manager of this hub.
+        /// </summary>
         public static async Task<IUserGroupManager> GetUserGroupManagerAsync(this InvocationContext invocationContext)
         {
             return (await StaticServiceHubContextStore.Get().GetAsync(invocationContext.Hub)).UserGroups;
--- a/src/SignalRServiceExtension/TriggerBindings/ServerlessHub.cs
+++ b/src/SignalRServiceExtension/TriggerBindings/ServerlessHub.cs
@@ -32,12 +32,24 @@ namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
             UserGroups = hubContext.UserGroups;
         }
 
+        /// <summary>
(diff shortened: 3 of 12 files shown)


src/SignalRServiceExtension/TriggerBindings/ServerlessHub.cs
```diff
@@ -30,6 +38,32 @@ public ServerlessHub()
 
         public IUserGroupManager UserGroups { get; }
 
+        public string HubName { get; }
```
JialinXin: Add summary for public properties.

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -80,8 +78,8 @@ public async Task LeaveUserFromGroup([SignalRTrigger]InvocationContext invocatio
             await UserGroups.RemoveFromGroupAsync(userName, groupName);
         }
 
-        [FunctionName(nameof(Disconnect))]
-        public void Disconnect([SignalRTrigger]InvocationContext invocationContext)
+        [FunctionName(nameof(OnDisconnect))]
+        public void OnDisconnect([SignalRTrigger]InvocationContext invocationContext)
```
wanlwanl: Why not OnDisconnected...? it doesn't consistent to `OnConnected`

src/SignalRServiceExtension/TriggerBindings/InvocationContextExtensions.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Azure.SignalR.Management;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    public static class InvocationContextExtensions
```
wanlwanl: add summary?

src/SignalRServiceExtension/TriggerBindings/SignalRFilterAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+            }
+            // Should not hit the Exception.
+            throw new InvalidOperationException($"{nameof(FunctionExceptionContext)} doesn't contain {nameof(InvocationContext)}.");
+        }
+
+        /// <summary>
+        /// Executed before the Function method being executed.
+        /// Throwing exceptions can terminate the Function execution and response the invocation failure.
+        /// </summary>
+        public abstract Task FilterAsync(InvocationContext invocationContext, CancellationToken cancellationToken);
```
wanlwanl: Can we make the filter returns true/false?

samples/bidirectional-chat/csharp/Function.cs
zackliu: Oh, nice catch.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -62,22 +62,29 @@ internal SignalRTriggerAttribute GetParameterResolvedAttribute(SignalRTriggerAtt
             if (declaredType != null && declaredType.IsSubclassOf(typeof(ServerlessHub)))
             {
                 // Class based model
+                if (!string.IsNullOrEmpty(hubName) ||
```
wanlwanl: make the properties' setters access level less than public?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -125,15 +132,32 @@ private void ValidateSignalRTriggerAttributeBinding(SignalRTriggerAttribute attr
 
         private string GetCategoryFromMethodName(string name)
         {
-            if (string.Equals(name, Event.Connected, StringComparison.OrdinalIgnoreCase) ||
-                string.Equals(name, Event.Disconnected, StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(name, Constants.OnConnected, StringComparison.OrdinalIgnoreCase) ||
+                string.Equals(name, Constants.OnDisconnected, StringComparison.OrdinalIgnoreCase))
```
JialinXin: hard code map to function name might be easily break and hard for customer to identify problem. Any way to notice the method name mapping logic?

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
wanlwanl: or move them to another class?

src/SignalRServiceExtension/TriggerBindings/SignalRFilterAttribute.cs
zackliu: The message of exception will be sent back to client side. So, I think let customer to throw exception maybe better. Or if they returns false, what should we throw?

src/SignalRServiceExtension/TriggerBindings/InvocationContextExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public static async Task<IHubClients> GetClientsAsync(this InvocationContext invocationContext)
```
wanlwanl: possilble to put Clients/Groups/UserGroups to InvocationContext?
Looks the Clients/Groups/UserGroups are only for C#, maybe add a new output of SignalRTriggerAttri that's only for C#?
Then the interface will be much similar to IHubContext

wanlwanl: Or add a new class implements IHubContext and use the InvocationContext as ctor param?

zackliu: I think in most C# cases, they should use class based model. Then they use that just like what they do in SDK. And for other few customers, we provided an extension to let them use it easily. It's not a common case, so as simple as this is OK I think.
Put them to InvocationContext is not good, and InvocationContext should keep simple as it passes between multiple language.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
zackliu: Removing Setters for these properties is good.

src/SignalRServiceExtension/TriggerBindings/SignalRFilterAttribute.cs
wanlwanl: I remember the exception throwed in function will respond 500 internal server error to client, so we can throw in OnExecutingAsync?

zackliu: No, it the behavior of httpTrigger. In SignalRTrigger, we handle the exception and send back the invocation response with message.

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -80,8 +78,8 @@ public async Task LeaveUserFromGroup([SignalRTrigger]InvocationContext invocatio
             await UserGroups.RemoveFromGroupAsync(userName, groupName);
         }
 
-        [FunctionName(nameof(Disconnect))]
-        public void Disconnect([SignalRTrigger]InvocationContext invocationContext)
+        [FunctionName(nameof(OnDisconnected))]
+        public void OnDisconnected([SignalRTrigger]InvocationContext invocationContext)
```
vicancy: is `OnDisconnectedAsync` a better matching? why `OnConnected` is Task `OnDisconnected` is void?

src/SignalRServiceExtension/TriggerBindings/SignalRFilterAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+#pragma warning disable CS0618 // Type or member is obsolete
+    public abstract class SignalRFilterAttribute : FunctionInvocationFilterAttribute
+    {
+        public override Task OnExecutingAsync(FunctionExecutingContext executingContext,
+            CancellationToken cancellationToken)
+        {
+            if (executingContext.Arguments.FirstOrDefault().Value is InvocationContext invocationContext)
```
vicancy: `FirstOrDefault()` can be `null`?

zackliu: It's a `KeyValuePair` and won't be null

samples/bidirectional-chat/csharp/Function.cs
zackliu: It's decided by customer, and we don't care about the return value. And the name need to be a convention, so I think the one without `Async` is enough.

src/SignalRServiceExtension/TriggerBindings/ServerlessHub.cs
```diff
@@ -2,34 +2,88 @@
...
+using System.Security.Claims;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Azure.SignalR.Management;
 
 namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
     /// <summary>
     /// When a class derived from <see cref="ServerlessHub"/>,
-    /// all the method in the are identified as using class based model.
+    /// all the method in the class are identified as using class based model.
```
JialinXin: typo: methods

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -12,25 +12,23 @@ namespace FunctionApp
...
-            [SignalRConnectionInfo(HubName = Hub, UserId = "{headers.x-ms-signalr-user-id}")] SignalRConnectionInfo connectionInfo)
+        public SignalRConnectionInfo Negotiate([HttpTrigger(AuthorizationLevel.Anonymous)]HttpRequest req)
         {
-            return connectionInfo;
+            return Negotiate(req.Headers["x-ms-signalr-user-id"], GetClaims(req.Headers["Authorization"]));
         }
 
-        [FunctionName(nameof(Connected))]
-        public async Task Connected([SignalRTrigger]InvocationContext invocationContext, ILogger logger)
+        [FunctionName(nameof(OnConnected))]
```
vwxyzh: function name should be constant

src/SignalRServiceExtension/TriggerBindings/ServerlessHub.cs
```diff
@@ -2,34 +2,88 @@
...
         private bool _disposed;
+        private readonly IServiceManager _serviceManager;
 
         public ServerlessHub()
         {
-            var hubName = GetType().Name;
-            var hubContext = StaticServiceHubContextStore.Get().GetAsync(hubName).GetAwaiter().GetResult();
+            HubName = GetType().Name;
+            var store = StaticServiceHubContextStore.Get();
+            var hubContext = store.GetAsync(HubName).GetAwaiter().GetResult();
```
vwxyzh: emm....

samples/bidirectional-chat/csharp/Function.cs
zackliu: Method name in class based model has special meanings. When method name changed, function name changes make sense. Constant is OK but `nameof` is also acceptable.

samples/bidirectional-chat/csharp/Function.cs
```diff
@@ -12,25 +12,23 @@ namespace FunctionApp
...
+        public async Task OnConnected([SignalRTrigger]InvocationContext invocationContext, ILogger logger)
```
chenkennt: How add `OnConnected` and `OnDisconnected` as virtual methods in base class?

zackliu: We don't know parameters of these two methods. Just like `Configure` and `ConfigureService` in `Startup`, which can only restrict by convention.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -62,22 +62,29 @@ internal SignalRTriggerAttribute GetParameterResolvedAttribute(SignalRTriggerAtt
             if (declaredType != null && declaredType.IsSubclassOf(typeof(ServerlessHub)))
             {
                 // Class based model
+                if (!string.IsNullOrEmpty(hubName) ||
+                    !string.IsNullOrEmpty(category) ||
+                    !string.IsNullOrEmpty(@event) ||
+                    parameterNames.Length != 0)
+                {
+                    throw new ArgumentException($"{nameof(SignalRTriggerAttribute)} must use No-Arg Constructor in class based model.");
```
chenkennt: In C# it's officially called *parameterless constructor*?

PR #110, opened 2020-03-19
Resolve attribute dynamically

wanlwanl:


src/SignalRServiceExtension/Bindings/SignalRInputBindings/AttributeCloner.cs
```diff
@@ -0,0 +1,532 @@
...
+
+using Validator = System.Action<object>;
+
+#pragma warning disable CS0618 // Type or member is obsolete
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    // Clone an attribute and resolve it.
+    // This can be tricky since some read-only properties are set via the constructor.
+    // This assumes that the property name matches the constructor argument name.
+    internal class AttributeCloner<TAttribute>
```
JialinXin: >internal class AttributeCloner<TAttribute> [](start = 4, length = 42)

Add comments about the file is copied from official external, so if further issue found we'd easier look for source. [https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs)

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRConnectionInputBinding.cs
```diff
@@ -1,38 +1,42 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+using System.Collections.Generic;
+using System.Reflection;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Protocols;
 using System.Threading.Tasks;
```
JialinXin: move system using before Microsoft, can add space line between system and rest.

src/SignalRServiceExtension/Bindings/SignalRInputBindings/InputBindingProvider.cs
```diff
@@ -29,10 +33,9 @@ public Task<IBinding> TryCreateAsync(BindingProviderContext context)
             {
                 switch (attr)
```
zackliu: Seems you can't get `GetCustomAttributes` from non-C# language. Thus, you will get a `null` to make the binding failed. Did you test in js?

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRConnectionInputBinding.cs
```diff
@@ -1,38 +1,42 @@
...
+using Microsoft.Extensions.Configuration;
 
 namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
-    internal class SignalRConnectionInputBinding : IBinding
+    internal class SignalRConnectionInputBinding : BindingBase<SignalRConnectionInfoAttribute>
```
zackliu: When binding the `SignalRConnectionInfo` to a parameter in non-C# language, Function tries to get the type of your binding data first. It will be `string` by default. Binding need to implement `IBindingRuleProvider` to provide the type of object. Did it work properly now in js?

wanlwanl: the `IBindingRuleProvider` is implemented in `FilteringBindingProvider`. the `InputBindingProvider` is an inner provider to `FilteringBindingProvider`. We don;t need to care about languages.

src/SignalRServiceExtension/Bindings/SignalRInputBindings/InputBindingProvider.cs
wanlwanl: Tested, works. It's an inner provider, the outer provider is in webjob sdk, we don't need to handle it.

wanlwanl: will support it later

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRConnectionInputBinding.cs
wanlwanl: will support it later

PR #113, opened 2020-03-22
Reduce header lookups in DefaultSecurityTokenValidator ValidateToken

danielwertheim:
As of my understandings, in general `TryGetValue` should be used instead of `ContainsKey` + `Value` lookup as it internally does a `TryGetValue`.

This PR reduces the number of accesses of `Authorization` header lookup and `ToString` calls to it. It also ignores casing of `Bearer` prefix. The current underlying implementation of `request.Headers` seem to be created ignoring-casing, so `authorization` vs `Authorization` should be safe.


wanlwanl:
Thanks for contributing!

PR #129, opened 2020-09-07
Fix SignalR Trigger makes whole function app down in Azure Static Web Apps

zackliu:
In Azure Static Web Apps, there's no storage account that required by webhook provider. As Azure Static Web Apps doesn't support all trigger's exception Http Trigger, we just disable SignalR Trigger in such case.

### DIFF
--- a/src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
+++ b/src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
@@ -80,10 +80,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
                 logger.LogInformation($"Registered SignalR trigger Endpoint = {url?.GetLeftPart(UriPartial.Path)}");
                 triggerEnabled = true;
             }
-            catch(Exception ex)
+            catch
             {
-                logger.LogWarning("SignalR trigger requires 'AzureWebJobsStorage' connection string being set. All SignalR trigger functions will be suppressed. " + 
-                    $"It's expected if you're using Azure Static Web Apps but not in other secnarios. {ex}");
+                logger.LogInformation("SignalR input and output bindings are enabled. " +
+                                      "To use the SignalR trigger, please set 'AzureWebJobsStorage' to a valid connection string.");
             }
             
             context.AddConverter<string, JObject>(JObject.FromObject)
--- a/src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
+++ b/src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
@@ -73,9 +73,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 
             StaticServiceHubContextStore.ServiceManagerStore = new ServiceManagerStore(options.AzureSignalRServiceTransportType, configuration, loggerFactory);
 
-            var url = context.GetWebhookHandler();
-            logger.LogInformation($"Registered SignalR trigger Endpoint = {url?.GetLeftPart(UriPartial.Path)}");
-
+            bool triggerEnabled = false;
+            try
+            {
+                var url = context.GetWebhookHandler();
+                logger.LogInformation($"Registered SignalR trigger Endpoint = {url?.GetLeftPart(UriPartial.Path)}");
+                triggerEnabled = true;
+            }
+            catch(Exception ex)
+            {
+                logger.LogWarning("SignalR trigger requires 'AzureWebJobsStorage' connection string being set. All SignalR trigger functions will be suppressed. " + 
+                    $"It's expected if you're using Azure Static Web Apps but not in other secnarios. {ex}");
+            }
+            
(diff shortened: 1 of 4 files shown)


zackliu:
Related PR: https://github.com/Azure/azure-functions-extension-bundles/pull/32

src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
```diff
@@ -73,9 +73,19 @@ public void Initialize(ExtensionConfigContext context)
...
+            try
+            {
+                var url = context.GetWebhookHandler();
+                logger.LogInformation($"Registered SignalR trigger Endpoint = {url?.GetLeftPart(UriPartial.Path)}");
+                triggerEnabled = true;
+            }
+            catch(Exception ex)
+            {
+                logger.LogWarning("SignalR trigger requires 'AzureWebJobsStorage' connection string being set. All SignalR trigger functions will be suppressed. " + 
+                    $"It's expected if you're using Azure Static Web Apps but not in other secnarios. {ex}");
```
anthonychu, line 86: I don't think this should be a warning with the exception displayed. I think the majority of customers will be using only the input/output bindings and don't need to have the storage connection string set (especially when running locally). They would be alarmed if they saw this message.

Perhaps display something like:

```suggestion
                logger.LogInformation("SignalR input and output bindings are enabled. " + 
                    "To use the SignalR trigger, please set 'AzureWebJobsStorage' to a valid connection string.");
```

Is it possible to test this build of the extension in Static Web Apps?

zackliu: From my experience, if you don't set `AzureWebJobsStorage` connection string, function app will directly crash with error (in local): `Missing value for AzureWebJobsStorage in local.settings.json. This is required for all triggers other than httptrigger, kafkatrigger. You can run 'func azure functionapp fetch-app-settings <functionAppName>' or specify a connection string in local.settings.json.` 
So, I think a warning makes sense because customer usually set `AzureWebJobsStorage` by default.

For the testing, does Static Web App support registering customer dependencies other than `bundles`. If it does, I can do a test

anthonychu: Checked Kusto and 40% of deployed function apps that use SignalR bindings have only http triggers (and therefore would get this warning if running without storage). That's why I'm concerned about how this message appears. Static web apps allow only HTTP triggers so their function apps fall into this category as well.

I'm actually not sure how to test this in Static Web Apps. Perhaps it's enough to run it in a regular function app with no storage account configured.

zackliu: Then I think log as information is enough. I've already tested it at where has no `AzureWebJobsStorage`

src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
```diff
@@ -73,9 +73,19 @@ public void Initialize(ExtensionConfigContext context)
...
+            try
+            {
+                var url = context.GetWebhookHandler();
+                logger.LogInformation($"Registered SignalR trigger Endpoint = {url?.GetLeftPart(UriPartial.Path)}");
+                triggerEnabled = true;
+            }
+            catch
+            {
+                logger.LogInformation("SignalR input and output bindings are enabled. " +
+                                      "To use the SignalR trigger, please set 'AzureWebJobsStorage' to a valid connection string.");
```
JialinXin, line 86: I think Log.Info only shows the information about important steps user hit. It seems not very effective to show suggestions in it. And if user doesn't use it correctly, the exception log throws is good enough to provide suggestions.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -39,6 +41,12 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
             {
                 return Task.FromResult<ITriggerBinding>(null);
             }
+
+            if (!_enabled)
```
JialinXin, line 45: Add a UT to cover the case to cover the negative case?

JialinXin (approved):

src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
zackliu: I save the exception and throw it only when customer try to use `SignalRTrigger`.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
zackliu: Added

zackliu:
@anthonychu  If you don't have more comments, I will release a new package and add trigger to bundle.

JialinXin (approved):

PR #149, opened 2020-12-01
 resolve service transport type in ServiceManagerStore

Y-Sindo:


src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
```diff
@@ -61,17 +64,7 @@ public void Initialize(ExtensionConfigContext context)
...
             }
 
-            var serviceTransportTypeStr = nameResolver.Resolve(Constants.ServiceTransportTypeName);
-            if (Enum.TryParse<ServiceTransportType>(serviceTransportTypeStr, out var transport))
-            {
-                options.AzureSignalRServiceTransportType = transport;
-            }
-            else
-            {
-                logger.LogWarning($"Unsupported service transport type: {serviceTransportTypeStr}. Use default {options.AzureSignalRServiceTransportType} instead.");
```
wanlwanl, line 71: what if the `serviceTransportTypeStr` is `trnsnt`

Y-Sindo: `System.InvalidOperationException : Unable to resolve service for type 'Microsoft.Azure.SignalR.Management.ServiceTransportType' while attempting to activate 'Microsoft.Azure.WebJobs.Extensions.SignalRService.ServiceManagerStore'.` will be thrown at https://github.com/Azure/azure-functions-signalrservice-extension/blob/bdba5d907925f4f71a8a054b3305d5c63cb9c96d/src/SignalRServiceExtension/Config/ServiceManagerStore.cs#L21

In fact, we are going to obsolete `ServiceTransportType` and might combine two implementation in the future.

wanlwanl: So you plan to throw exception in the future while the current one returns default value?

Y-Sindo: use default value now.

wanlwanl (approved):

Issue #148, opened 2020-11-13
SignalR Trigger doesn't support duplicated claims


zackliu:
As claims are represented as dictionary, it doesn't support duplicated key. If client invoke with duplicated claims, 500 will be returned.

https://github.com/Azure/azure-functions-signalrservice-extension/blob/30a08f2d96b87a6fd22f4d9507807734256dffef/src/SignalRServiceExtension/TriggerBindings/Resolver/SignalRRequestResolver.cs#L88

PR #151, opened 2020-12-08
Support duplicated key in query, header, claims

zackliu:
* For headers and queries, we join values with the same key by comma
* For claims, we reserve the first claim when multiple claims have the same key

- #148


JialinXin (approved):
:shipit:

PR #153, opened 2020-12-10
Remove `AppSetting` attribute of `ConnectionStringSetting` of trigger…

Y-Sindo:
… binding

* remove SignalROptions


src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -99,41 +101,21 @@ internal SignalRTriggerAttribute GetParameterResolvedAttribute(SignalRTriggerAtt
                 ? parameterNamesFromAttribute
                 : parameterNames;
 
-            var resolvedConnectionString = GetResolvedConnectionString(
-                typeof(SignalRTriggerAttribute).GetProperty(nameof(attribute.ConnectionStringSetting)),
-                attribute.ConnectionStringSetting);
-
-            return new SignalRTriggerAttribute(hubName, category, @event, parameterNames) {ConnectionStringSetting = resolvedConnectionString};
+            return new SignalRTriggerAttribute(hubName, category, @event, parameterNames) { ConnectionStringSetting = attribute.ConnectionStringSetting };
```
zackliu, line 104: Any reason to move the resolved connection string to a separate parameter of `SignalRTriggerBinding` instead of put it into `SignalRTriggerAttribute`

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -99,41 +101,21 @@ internal SignalRTriggerAttribute GetParameterResolvedAttribute(SignalRTriggerAtt
...
-                {
-                    throw new InvalidOperationException($"Unable to get AppSettingAttribute on property {property.Name}");
-                }
-                resolvedConnectionString = _nameResolver.Resolve(attribute.Default);
-            }
-
-            return string.IsNullOrEmpty(resolvedConnectionString)
-                ? _options.ConnectionString
-                : resolvedConnectionString;
+            var monitor = (_managerStore.GetOrAddByConnectionStringKey(connectionStringKey) as ServiceHubContextStore).OptionsMonitor;
```
zackliu, line 109: I prefer add a `ConnectionString` property to `IServiceHubContextStore`. It will make this more general, especially in test. As the `_managerStore` is a interface, we can mock it and generate a mocked `IServiceHubContextStore` in test.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
Y-Sindo: In the multiple-endpoint scenario, there are multiple connection strings and we need to resolve them with customized method in SDK. So resolving only one connection string outside SDK will be meaningless. The connection string parameter of  `SignalRTriggerBinding` will be replaced by connection string array or access key array in the future.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
Y-Sindo: We want dynamic configuration so we need `IOptionsMonitor` to get the runtime value of `ServiceManagerOptions`. How about add `IOptionsMonitor` as an internal get-only property to `IServiceHubContextStore`?

zackliu: What about adding `public string ConnectionString {get;}` to interface and implement `public string ConnectionString => optionsMonitor.CurrentValue.ConnectionString`. Thus you can still get the latest value. Seems exposing the whole IOptionsMonitor is not necessary.

zackliu (approved):

PR #157, opened 2021-01-28
Support languages other than C# to specify service endpoints to send message

Y-Sindo:


src/SignalRServiceExtension/Bindings/SignalRData.cs
```diff
@@ -1,11 +1,14 @@
...
 
+using Microsoft.Azure.SignalR;
+
 namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
     internal class SignalRData
     {
         public string Target { get; set; }
         public object[] Arguments { get; set; }
+        public ServiceEndpoint[] Endpoints { get; set; }
```
JialinXin, line 12: `ServiceEndpoint` is an class we defined in c#, but not recognized in non c# language, like js, python, java. Usually built-in types are safe to use, like string, int, arrays. Shall we make this string[] and do some further process later.

PR #163, opened 2021-02-08
Support Function V3 runtime

Y-Sindo:


test/SignalRServiceExtension.Tests/DefaultSecurityTokenValidatorTests.cs
```diff
@@ -40,7 +38,7 @@ public class DefaultSecurityTokenValidatorTests
         public void ValidateSecurityTokenFacts(string tokenString, SecurityTokenStatus expectedStatus)
         {
             var ctx = new DefaultHttpContext();
-            var req = new DefaultHttpRequest(ctx);
+            var req = ctx.Request;
```
Y-Sindo, line 41: NetCore 3.0 doesn't support `new DefaultHttpRequest(ctx)`

Issue #156, opened 2020-12-29
Unsupported service transport type


gavinsteinhoff:
I keep getting this message when my C# function app is starting.

> Unsupported service transport type: . Use default Transient instead.

I have a basic function app setup to use signalR:

```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
	  <TargetFramework>netcoreapp3.1</TargetFramework>
	  <AzureFunctionsVersion>v3</AzureFunctionsVersion>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.2.5" />
    <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
    <PackageReference Include="Microsoft.Extensions.Options.DataAnnotations" Version="5.0.0" />
    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.11" />
    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.SignalRService" Version="1.2.2" />
  </ItemGroup>
```
I can still negotiate with the signalR service, but I would like to figure out why this message is here.

```
Azure Functions Core Tools
Core Tools Version:       3.0.3160 Commit hash: 00aa7f43cc5c5f15241b5e6e5363256f19ceb990
Function Runtime Version: 3.0.14916.0
...
[2020-12-29T16:30:41.028Z] Services registered by external startup type Microsoft.Azure.WebJobs.Extensions.SignalRService.SignalRWebJobsStartup:
[2020-12-29T16:30:41.031Z]  Microsoft.Azure.WebJobs.Host.Config.IExtensionConfigProvider: Implementation: Microsoft.Azure.WebJobs.Extensions.SignalRService.SignalRConfigProvider, Lifetime: Singleton
[2020-12-29T16:30:41.033Z]  Microsoft.Extensions.Options.IConfigureOptions`1[[Microsoft.Azure.WebJobs.Extensions.SignalRService.SignalROptions, Microsoft.Azure.WebJobs.Extensions.SignalRService, Version=1.2.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60]]: Factory, Lifetime: Singleton
[2020-12-29T16:30:41.227Z] Active host changing from '(null)' to 'b6f82d15-61e0-443a-a55e-7c4f36fdcc3f'.
[...]

PR #164, opened 2021-02-09
Clarify missing AzureSignalRServiceTransportType error

bmc-msft:
When AzureSignalRServiceTransportType is unset, users get the log message `Unsupported service transport type: . Use default Transient instead.` which is confusing.

This clarifies it's using the default setting if not set.


bmc-msft:
This is related to #156

JialinXin (approved):

PR #187, opened 2021-03-18
Add hub protocol configuration on netcore3.1

Y-Sindo:
The change that making newtonsoft the defalut protocol on netcore 3.1 is a breaking change.
Now restore the defalut protocol on netcore 3.1 to System.Text.Json and allow user to use `Azure:SignalR:HubProtocol` to configure it.

### DIFF
--- a/src/SignalRServiceExtension/Constants.cs
+++ b/src/SignalRServiceExtension/Constants.cs
@@ -7,6 +7,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
     {
         public const string AzureSignalRConnectionStringName = "AzureSignalRConnectionString";
         public const string AzureSignalREndpoints = "Azure:SignalR:Endpoints";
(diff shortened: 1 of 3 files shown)


src/SignalRServiceExtension/Constants.cs
```diff
@@ -7,6 +7,7 @@ internal static class Constants
     {
         public const string AzureSignalRConnectionStringName = "AzureSignalRConnectionString";
         public const string AzureSignalREndpoints = "Azure:SignalR:Endpoints";
+        public const string AzureSignalRHubProtocol = "Azure:SignalR:HubProtocol";
```
wanlwanl, line 10: if someone use standard2, then it doesn't work

wanlwanl: can this protocol config also apply to netstandard2.0?

Y-Sindo: This protocol config is designed for netcore3.1

wanlwanl (approved):

PR #196, opened 2021-03-29
Add Multi ConnectionInfo for external negotiation routing

Y-Sindo:
# Usage sample

```json
    {
      "type": "SignalRNegotiation",
      "name": "negotiateCtx",
      "userId": "{headers.x-ms-signalr-userid}",
      "hubName": "simplechat",
      "direction": "in"
    }
```

```js
module.exports = function (context, req, negotiateCtx) {
  console.log(context);
  context.res = { body: negotiateCtx.endpoints.find( e=>e.online ).connectionInfo };
  context.done();
};
```


test/SignalRServiceExtension.Tests/AzureSignalRClientTests.cs
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+// Copyright (c) Microsoft. All rights reserved.
```
wanlwanl, line 4: ?

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.SignalR;
+using Microsoft.Azure.SignalR.Management;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    internal class MultiConnectionInfoAsyncConverter : IAsyncConverter<SignalRMultiConnectionInfoAttribute, Dictionary<ServiceEndpoint, SignalRConnectionInfo>>
```
wanlwanl, line 13: @vicancy doesn't like abbr. in naming

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+                .GetAsync(input.HubName) as IInternalServiceHubContext;
+            var endpoints = serviceHubContext.GetServiceEndpoints();
+            var dict = endpoints.ToDictionary(e => e, async e =>
+             {
+                 var subHubContext = serviceHubContext.WithEndpoints(new ServiceEndpoint[] { e });
+                 var azureSignalRClient = new AzureSignalRClient(subHubContext);
+                 var negotiationRes = await azureSignalRClient.GetClientConnectionInfoAsync(input.UserId, input.IdToken, input.ClaimTypeList, null);
+                 return negotiationRes;
+             });
+            _ = await Task.WhenAll(dict.Values);
```
wanlwanl, line 36: remove `_ =`

test/SignalRServiceExtension.Tests/MultiConnectionInfoAsyncConverterTest.cs
```diff
@@ -0,0 +1,89 @@
...
+            var serviceManagerStore = new ServiceManagerStore(configuration, NullLoggerFactory.Instance);
+            var converter = new MultiConnectionInfoAsyncConverter(serviceManagerStore);
+
+            var userId = "User";
+            var idToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c";
+            var expectedName = "John Doe";
+            var expectedIat = "1516239022";
+            var claimTypeList = new string[] { "name", "iat" };
+            var attribute = new SignalRMultiConnectionInfoAttribute 
+            { HubName = HubName,
```
wanlwanl, line 60: move to next line

test/SignalRServiceExtension.Tests/MultiConnectionInfoAsyncConverterTest.cs
```diff
@@ -0,0 +1,89 @@
...
+              UserId = userId, 
+              IdToken = idToken,
+              ClaimTypeList = claimTypeList
+            };
+            var dict = await converter.ConvertAsync(attribute, default);
+            var (endpoint, connectionInfo) = dict.First();
+            Assert.Equal(connectionInfo.Url, $"{endpoint.Endpoint}/client/?hub={HubName.ToLower()}");
+            var claims = new JwtSecurityTokenHandler().ReadJwtToken(connectionInfo.AccessToken).Claims;
+            Assert.Equal(expectedName, claims.First(c => c.Type.Equals("name")).Value);
+            Assert.Equal(expectedIat, claims.First(c => c.Type.Equals($"{AzureSignalRClient.AzureSignalRUserPrefix}iat")).Value);
```
wanlwanl, line 70: what about IdToken, ClaimTypeList? why set them if you don't validate?

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+    internal class MultiConnectionInfoAsyncConverter : IAsyncConverter<SignalRMultiConnectionInfoAttribute, Dictionary<ServiceEndpoint, SignalRConnectionInfo>>
+    {
+        private readonly IServiceManagerStore _serviceManagerStore;
+
+        public MultiConnectionInfoAsyncConverter(IServiceManagerStore serviceManagerStore)
+        {
+            _serviceManagerStore = serviceManagerStore;
+        }
+
+        public async Task<Dictionary<ServiceEndpoint, SignalRConnectionInfo>> ConvertAsync(
```
wanlwanl, line 22: cx get `Dictionary<ServiceEndpoint, SignalRConnectionInfo>`, so the router need to be outside of extension? why you don't use the endpoint router and then return a `SignalRConnectionInfo` so that cx can simply return it.

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            var serviceHubContext = await _serviceManagerStore
+                .GetOrAddByConnectionStringKey(input.ConnectionStringSetting)
+                .GetAsync(input.HubName) as IInternalServiceHubContext;
+            var endpoints = serviceHubContext.GetServiceEndpoints();
+            var dict = endpoints.ToDictionary(e => e, async e =>
+             {
+                 var subHubContext = serviceHubContext.WithEndpoints(new ServiceEndpoint[] { e });
+                 var azureSignalRClient = new AzureSignalRClient(subHubContext);
+                 var negotiationRes = await azureSignalRClient.GetClientConnectionInfoAsync(input.UserId, input.IdToken, input.ClaimTypeList, null);
```
wanlwanl, line 33: I don't think it's necessary to negotiate with all the endpoints

src/SignalRServiceExtension/SignalRMultiConnectionInfoAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Description;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    [AttributeUsage(AttributeTargets.ReturnValue | AttributeTargets.Parameter)]
+    [Binding]
+    public class SignalRMultiConnectionInfoAttribute : Attribute
```
wanlwanl, line 11: maybe discuss the naming later. already use `Info`... but we should not use abbr in naming for public usage (by @vicancy )

wanlwanl, line 11: add descriptions. ///summary

test/SignalRServiceExtension.Tests/MultiConnectionInfoAsyncConverterTest.cs
wanlwanl: since you already need to read token, why don't you validate all the components of multiple  you want to validate in a `[Theory]`

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
wanlwanl: oh, I see, it's mainly for js, right?

Y-Sindo: yes

test/SignalRServiceExtension.Tests/MultiConnectionInfoAsyncConverterTest.cs
Y-Sindo: Just follow the test in `AzureSignalRClientTests.GetClientConnectionInfo`

src/SignalRServiceExtension/SignalRMultiConnectionInfoAttribute.cs
Y-Sindo: `SignalRConnectionInfoAttribute` doesn't have descriptions too.

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
Y-Sindo: This is suggested by IDE .

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
Y-Sindo: We already have `MultiEndpointServiceConnectionContainer` in SDK? If we don't use abbr. , then the name will be super long. I don't think this is a good experience for users and us.

test/SignalRServiceExtension.Tests/MultiConnectionInfoAsyncConverterTest.cs
Y-Sindo: > since you already need to read token, why don't you validate all the components of multiple you want to validate in a `[Theory]`

What do you mean by "validate all the components of multiple"?

src/SignalRServiceExtension/SignalRMultiConnectionInfoAttribute.cs
vicancy: The properties inside the class do not show anything related to "multi" though..

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
vicancy: maybe some `ConnectionInfoMap` instead of `Multi...`

src/SignalRServiceExtension/SignalRMultiConnectionInfoAttribute.cs
wanlwanl: still need to add it.

test/SignalRServiceExtension.Tests/MultiConnectionInfoAsyncConverterTest.cs
wanlwanl: oh, I see, the claims are validated, just forget take a look at what's inside idtoken. the test is ok.
suggest to remove it, since it already covered. this test isn't about multi endpoint.

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
Y-Sindo: I found using a endpoint object as the map key is not convenient to convert the map to JSON. Intend to change the map to a array like this:
```json
[
    {
          "endpoints":{},
          "connectionInfo":{}
    },
]
```
Giving this, we can still use `MultiConnectionInfo`?

vicancy: > endpoint object as the map key is not convenient to convert the map to JSON

where is inconvenient?

vicancy: A potential issue for using array is array is not extensible for adding more metadata properties, so generally we would prevent using array as the top level type, is this JSON array the top type?

src/SignalRServiceExtension/SignalRMultiConnectionInfoAttribute.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    /// <summary>
+    /// An input binding attribute to bind a list of <see cref="EndpointConnectionInfo"/> to the function parameter.
+    /// </summary>
+    /// <remarks>Designed for function languages except C# to customize negotiation routing.</remarks>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    [Binding]
+    public class SignalRConnectionInfoListAttribute : Attribute
```
vicancy, line 15: what's the difference between this one and the `SignalRConnectionInfoAttribute`, they looks pretty similar

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
wanlwanl: what about `ConnectionInfos` or `NegotiationResponses`?

src/SignalRServiceExtension/SignalRMultiConnectionInfoAttribute.cs
Y-Sindo: `SignalRConnectionInfoListAttribute` binds multiple connection info objects of all endpoints to the parameter, `SignalRConnectionInfoAttribute`  binds one connection info objects for a endpoint selected by endpoint router. This two bindings just have same input.

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
Y-Sindo: How about `NegotiateContext`

wanlwanl: what about reuse the standard one https://github.com/dotnet/aspnetcore/blob/main/src/SignalR/common/Http.Connections.Common/src/NegotiationResponse.cs , signalr users should have knowledge about it @vicancy

Y-Sindo: JSON can not represent a map whose keys are object.

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/NegotiateContext.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    internal class NegotiateContext
+    {
+        public EndpointConnectionInfo[] ClientEndpoints { get; set; }
```
vicancy, line 8: how about `Endpoints`?

src/SignalRServiceExtension/Bindings/EndpointConnectionInfo.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.Azure.SignalR;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    [JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
+    internal class EndpointConnectionInfo
+    {
+        public ServiceEndpoint ServiceEndpoint { get; set; }
```
vicancy, line 13: how about `EndpointConnectionInfo: ServiceEndpoint`

src/SignalRServiceExtension/Bindings/EndpointConnectionInfo.cs
```diff
@@ -0,0 +1,17 @@
...
+        public string Url { get; set; }
+        public string AccessToken { get; set; }
```
wanlwanl, line 15: is it auto generate josn object to for js cx? 
then if cx returns this to client, the client will see connection string? since `ServiceEndpoint` contains it.
or we can control it somewhere?

wanlwanl: looks like if this instance is returned to client in response, function app admin need to generate a new object only contains `url` and `accessToken` to avoid secret leak. then function app admin have to have the knowledge about negotiation response, and ask us what to set about `ConnectionId` and `AvailableTransports`. maybe a property only for  `url` and `accessToken`  is fine, admin can simply respond it. @vicancy @Y-Sindo

Y-Sindo: Not all of the properties in `ServiceEndpoint` will be serialized, see https://github.com/Azure/azure-functions-signalrservice-extension/blob/3f33c997848ace6d5d6b2c2ca11195168a8f92d7/src/SignalRServiceExtension/Bindings/ServiceEndpointJsonConverter.cs#L10 . 
But I think we can consider whether wrapping url and accessToken in a separate object would be convenient for users.

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/NegotiateContext.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    internal class NegotiateContext
```
wanlwanl, line 6: nit: naming: `Negotiate` -> `Negotiation`

src/SignalRServiceExtension/NegotiateBaseAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Description;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    [AttributeUsage(AttributeTargets.ReturnValue | AttributeTargets.Parameter)]
+    [Binding]
+    public class NegotiateBaseAttribute : Attribute
```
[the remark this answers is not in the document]
wanlwanl: description?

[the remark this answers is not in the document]
wanlwanl: public ->internal?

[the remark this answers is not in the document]
wanlwanl: It's public, If cx use this attribute by accident, we don't have a binding to handle it.

src/SignalRServiceExtension/Bindings/SignalRInputBindings/SignalRMultiConnectionInfoBinding/MultiConnectionInfoAsyncConverter.cs
wanlwanl: about the naming, since we already has `SignalRConnectionInfo`, and now we decide to returns a list, how about naming the return type as `SignalRConnectionInfoList` and `[SignalRConnectionInfoList]`? @JialinXin

wanlwanl: > I found using a endpoint object as the map key is not convenient to convert the map to JSON
how about using endpoint name?
Consider "Resiliency and disaster recovery", they already has the key as `primary` and `secondary`, they try to get the primary first, if its offline, then to the secondary. So a dictionary is more convenient.

Y-Sindo: What is the type of es?

Y-Sindo: "they already has the key as primary and secondary, they try to get the primary first, if its offline, then to the secondary. So a dictionary is more convenient. " 
I am not sure I understand your meaning. Do you mean that the user already has a list of names of primary endpoints , and they will loop the dictionary by the names? But I don't think user would maintain a list of names of endpoints. What's more, a dictionary is not convenient for users to get a random endpoint.

Y-Sindo: I just find that since ES6, we can loop over object by Object.Keys(), so we could consider it.

wanlwanl: yeah.

Y-Sindo: But SDK just return a list to user. Using a list is more consistent with SDK.

src/SignalRServiceExtension/NegotiationContextAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+    /// <summary>
+    /// An input binding attribute to bind <see cref="NegotiationContext"/> to the function parameter.
+    /// </summary>
+    /// <remarks>Designed for function languages except C# to customize negotiation routing.</remarks>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    [Binding]
+    public class SignalRNegotiationAttribute : NegotiationBaseAttribute
+    {
```
Y-Sindo, line 16: What do you think about `SignalRNegotiationAttribute` instead of `SignalRNegotiationContext`? As `HttpTrigger` doesn't use `HttpResponse` as the attribute name, so I think use `SignalRNegotiation` would be more concise.

vicancy: Looks good to me

vicancy (approved):

JialinXin (approved):

PR #198, opened 2021-04-01
Automatically switch to Newtonsoft for Non CSharp language worker

Y-Sindo:


src/SignalRServiceExtension/Config/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+#endif
+            }
+            return services;
+        }
+
+#if NETCOREAPP3_1
+        private static bool Not_DotnetRuntime(IConfiguration configuration)
+        {
+            var workerRuntime = configuration[Constants.FunctionsWorkerRuntime];
+            return workerRuntime!=null && !configuration[Constants.FunctionsWorkerRuntime].Equals(Constants.DotnetWorker);
```
JialinXin, line 43: miss space -> workerRuntime != null

src/SignalRServiceExtension/Config/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+            if (Environment.Version.Major == 4)
+            {
+                // Actually is .Net Core 2.x
+                if (configuration[Constants.AzureSignalRHubProtocol] != null)
+                {
+                    throw new InvalidOperationException(HubProtocolError);
+                }
+            }
+            else
+            {
```
JialinXin, line 27: remove the else?

src/SignalRServiceExtension/Config/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+#if NETCOREAPP3_1
```
JialinXin, line 28: missed 3.0 and 5.0+?

src/SignalRServiceExtension/Config/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+                {
+                    //this would overwrite the System.Text.Json Protocol.
+                    services.TryAddEnumerable(ServiceDescriptor.Singleton<IHubProtocol, NewtonsoftJsonHubProtocol>());
+                }
+#endif
+            }
+            return services;
+        }
+
+#if NETCOREAPP3_1
```
JialinXin, line 39: remove the framework check? too much if.

JialinXin, line 39: remove the framework check? too much if.

src/SignalRServiceExtension/Config/DependencyInjectionExtensions.cs
JialinXin, line 28: missed 3.0 and 5.0+?

Y-Sindo: Currently Azure Function only have 2.2 and 3.1 runtime.

src/SignalRServiceExtension/Config/DependencyInjectionExtensions.cs
Y-Sindo: `NewtonsoftJsonHubProtocol` only exists in .Net 3.0 +

src/SignalRServiceExtension/Config/DependencyInjectionExtensions.cs
JialinXin: I see. So for 2.2 how about use NETCOREAPP2_2 instead of Environment major version looks tricky. Beside, can `#if` check can replace logical if check?

src/SignalRServiceExtension/Config/DependencyInjectionExtensions.cs
JialinXin: how about move this simple method into above function, to reduce `#if`?

src/SignalRServiceExtension/Config/DependencyInjectionExtensions.cs
Y-Sindo: Environment major version is more robust, because if uses set their target framework to netstandard 2.0 and run their function on netcoreapp2.2, then NETCOREAPP2_2 would not work.

src/SignalRServiceExtension/Config/DependencyInjectionExtensions.cs
Y-Sindo: Then it would become:
```cs
var userSpecifyNewtonsoft = configuration.GetValue(Constants.AzureSignalRHubProtocol, HubProtocol.SystemTextJson) == HubProtocol.NewtonsoftJson;

if (Not_DotnetRuntime(configuration) || userSpecifyNewtonsoft )
                {
                    //this would overwrite the System.Text.Json Protocol.
                    services.TryAddEnumerable(ServiceDescriptor.Singleton<IHubProtocol, NewtonsoftJsonHubProtocol>());
                }
```
Seems not symmetrical and less readability? Also no need to compile the code in the second `#if` .

JialinXin (approved):

PR #205, opened 2021-04-09
Fix bidirectional-chat error

Y-Sindo:
* Use async negotiation method
* Remove the get `index.html` method as it doesn't work on Azure now.
* Add two scripts to start a http server to serve `index.html`.
* Modify README.md
* Move static files such png files, html file to `bidirectional-chat/content` directory.
* Rename `extensions.csproj` to `bidirectional-chat.csproj`, add `bidirectional-chat.sln` file, so that we can find the project in the Visual Studio solution view.


zackliu:
What about include the project to `azure-functions-signalrservice-extension.sln`. Thus, we don't need two `sln`.

Y-Sindo:
> What about include the project to `azure-functions-signalrservice-extension.sln`. Thus, we don't need two `sln`.

But other sample projects have their own solutions.

JialinXin:
Not much need to include a new sln. Either make all samples available in the main solution or leave sample proj is enough. sln can be created on demand.

JialinXin (approved):

PR #233, opened 2021-05-20
Fix override global default json serialization options

### LINKED ISSUE #229: Use of JsonConvert.DefaultSettings breaks unrelated serialization in v1.4
Upgrading from version 1.3.0 to 1.4.0 resulted in breaking changes for other aspects of my project. The project uses JsonConvert.SerializeObject() with default settings, which retains the POCO property casing (PascalCase in my situation). After the 1.4 update JsonConvert.SerializeObject() defaults to camelCase. This is running locally via Visual Studio 2019, I'm not sure if the same behavior exists deployed to Azure infrastructure.

I assume it's due to the following code fragment in SignalRConfigProvider.Initialize(), which is marked [Obsolete("preview")].

https://github.com/Azure/azure-functions-signalrservice-extension/blob/dbe3704a6e45f0624183d4152296197fad7ec253/src/SignalRServiceExtension/Config/SignalRConfigProvider.cs#L68-L75

I'd say it's bad practice to set global serialization defaults in a library like this, that should be reserved for top level applications. Custom JsonSerializerSettings should be passed via the appropriate overload to JsonConvert methods.

Y-Sindo:
fix #229 
Already passed the E2E tests.

### DIFF
--- a/src/SignalRServiceExtension/Bindings/EndpointConnectionInfo.cs
+++ b/src/SignalRServiceExtension/Bindings/EndpointConnectionInfo.cs
@@ -3,10 +3,11 @@
 
 using Microsoft.Azure.SignalR;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
-    [JsonObject]
+    [JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
     internal class EndpointConnectionInfo : LiteServiceEndpoint
     {
         public EndpointConnectionInfo(ServiceEndpoint endpoint) : base(endpoint)
--- a/src/SignalRServiceExtension/Bindings/JTokenExtensions.cs
+++ b/src/SignalRServiceExtension/Bindings/JTokenExtensions.cs
@@ -8,16 +8,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
     internal static class JTokenExtensions
     {
-        public static readonly JsonSerializer JsonSerializer = JsonSerializer.Create(new JsonSerializerSettings
-        {
-            Converters = new JsonConverter[] { new ServiceEndpointJsonConverter() }
-        });
-
         public static bool TryToObject<TOutput>(this JToken input, out TOutput output)
         {
             try
             {
-                output = input.ToObject<TOutput>(JsonSerializer);
+                output = input.ToObject<TOutput>(ServiceEndpointJsonConverter.JsonSerializer);
             }
             catch
             {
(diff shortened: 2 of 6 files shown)


wanlwanl:
Is it a fix to some issue? If so, please add the issue link here

src/SignalRServiceExtension/Bindings/EndpointConnectionInfo.cs
```diff
@@ -3,10 +3,11 @@
 
 using Microsoft.Azure.SignalR;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
-    [JsonObject]
+    [JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
```
wanlwanl, line 10: why `CamelCaseNamingStrategy` is required? the default one doesn't work?

Y-Sindo: Newtonsoft is not camel case by default?

Y-Sindo:
> Is it a fix to some issue? If so, please add the issue link here

Yes, already referred to the issue now in the PR description now.

src/SignalRServiceExtension/Bindings/JTokenExtensions.cs
```diff
@@ -1,17 +1,23 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
 {
     internal static class JTokenExtensions
     {
+        public static readonly JsonSerializer JsonSerializer = JsonSerializer.Create(new JsonSerializerSettings
```
JialinXin, line 11: How about move all the customize JsonConverter to `ConfigProvider` as global settings, so you don't have to config it specifically. 

Something like this: https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/webpubsub/Microsoft.Azure.WebJobs.Extensions.WebPubSub/src/Config/WebPubSubConfigProvider.cs#L150

src/SignalRServiceExtension/Config/SignalRConfigProvider.cs
```diff
@@ -96,7 +86,8 @@ public void Initialize(ExtensionConfigContext context)
                 .BindToInput(new NegotiationContextAsyncConverter(serviceManagerStore));
 
             _ = context.AddBindingRule<SignalREndpointsAttribute>()
-                   .AddConverter<ServiceEndpoint[], JArray>(JArray.FromObject)
+                   .AddConverter<ServiceEndpoint[], JArray>(
+                    endpoints => JArray.FromObject(endpoints, JTokenExtensions.JsonSerializer))
```
JialinXin, line 90: add a extra tab before `endpoints`?

src/SignalRServiceExtension/Bindings/JTokenExtensions.cs
Y-Sindo: I did it the same before... but reported from customer that this setting overrides the customer's setting. So I have made this PR. See #229

JialinXin (approved):

PR #248, opened 2021-06-23
Add Newtonsoft CamelCase config support

### LINKED ISSUE #209: camelCase serialisation in output binding.
https://docs.microsoft.com/en-us/aspnet/core/signalr/configuration?view=aspnetcore-5.0&tabs=dotnet#jsonmessagepack-serialization-options

If self hosting then it defaults to using camelCase, but it appears that the azure function version requires us to manually serialise the message ourselves.

Can either the default behaviour be changed to match the normal client/server defaults or some form of configuration added to enable specifying the behaviour?

Y-Sindo:
Allow to use "Azure:SignalR:HubProtocol:NewtonsoftJson:CamelCase=true" to enable camel case.
Fixed #209


src/SignalRServiceExtension/Constants.cs
```diff
@@ -8,6 +8,7 @@ internal static class Constants
         public const string AzureSignalRConnectionStringName = "AzureSignalRConnectionString";
         public const string AzureSignalREndpoints = "Azure:SignalR:Endpoints";
         public const string AzureSignalRHubProtocol = "Azure:SignalR:HubProtocol";
+        public const string AzureSignalRNewtonsoftCamelCase = "Azure:SignalR:HubProtocol:Newtonsoft:CamelCase";
```
JialinXin, line 11: Would this config too deep?

JialinXin: `Azure:SignalR:HubProtocol` accepts `SystemJson` and `NewtonsoftJson`, is there relationship with this config?

Y-Sindo: `AzureSignalRNewtonsoftCamelCase` would take effect only when `Azure:SignalR:HubProtocol=NewtonsoftJson`.

src/SignalRServiceExtension/Constants.cs
```diff
@@ -8,6 +8,7 @@ internal static class Constants
         public const string AzureSignalRConnectionStringName = "AzureSignalRConnectionString";
         public const string AzureSignalREndpoints = "Azure:SignalR:Endpoints";
         public const string AzureSignalRHubProtocol = "Azure:SignalR:HubProtocol";
+        public static string AzureSignalRNewtonsoftCamelCase = $"{AzureSignalRNewtonsoftCamelCase}:{HubProtocol.NewtonsoftJson}:CamelCase";
```
JialinXin, line 11: Typo. Update tests for better coverage.
> $"{AzureSignalRHubProtocol}:{HubProtocol.NewtonsoftJson}:CamelCase";

JialinXin (approved):

JialinXin (approved):

PR #263, opened 2021-08-14
Hub Context Attribute

phantomcosmonaut:
allows developers to set a custom connection string setting name that will be passed to the hub context property. The attribute is picked up by trigger bindings automatically


Y-Sindo:
How about letting the `ServerlessHub` virtually implements `IConnectionProvider` ?

phantomcosmonaut:
> How about letting the `ServerlessHub` virtually implements `IConnectionProvider` ?

OK, I did this but not virtually. Not sure how to resolve connection string within `SignalRTriggerAttribute` now, but I am fine with explicitly setting it.

src/SignalRServiceExtension/TriggerBindings/ServerlessHub.cs
```diff
@@ -41,15 +41,20 @@ public abstract class ServerlessHub : IDisposable
...
             Groups = hubContext.Groups;
             UserGroups = hubContext.UserGroups;
             _hubContext = hubContext as ServiceHubContext;
             ClientManager = _hubContext?.ClientManager;
         }
 
+        /// <summary>
+        /// The connection string setting name.
+        /// </summary>
+        public string Connection { get; set; } = Constants.AzureSignalRConnectionStringName;
```
phantomcosmonaut, line 56: Feels like this should be `get` only, but then it doesn't implement `IConnectionProvider`, but not sure how important that is

Y-Sindo: I think it is OK.

src/SignalRServiceExtension/TriggerBindings/SignalRTriggerBindingProvider.cs
```diff
@@ -52,9 +52,9 @@ public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext
             var resolvedAttribute = GetParameterResolvedAttribute(attribute, parameterInfo);
             ValidateSignalRTriggerAttributeBinding(resolvedAttribute);
 
-            var accessKeys = _managerStore.GetOrAddByConnectionStringKey(attribute.ConnectionStringSetting).AccessKeys;
+            var accessKeys = _managerStore.GetOrAddByConnectionStringKey(resolvedAttribute.ConnectionStringSetting).AccessKeys;
```
Y-Sindo, line 55: I think no need to change this. We want to original value of `ConnectionStringSetting` instead of the resolved connection string. Actually the `ConnectionStringSetting` is not an `[AutoResolve]` attribute.

src/SignalRServiceExtension/Config/OptionsSetup.cs
```diff
@@ -39,7 +39,7 @@ public OptionsSetup(IConfiguration configuration, ILoggerFactory loggerFactory,
 
         public void Configure(ServiceManagerOptions options)
         {
-            options.ConnectionString = configuration[connectionStringKey];
+            options.ConnectionString = configuration.GetConnectionString(connectionStringKey) ?? configuration[connectionStringKey];
```
Y-Sindo, line 42: Seems no related to the PR. And we don't want to support too many configuration styles, otherwise it is a little messy.

Y-Sindo:
Sorry, I found we can't get the `Connection` field of `ServerlessHub` from `SignalRTriggerBindingProvider`, so we have to switch back to attribute style. 

I created a branch and here is the sample : https://github.com/Azure/azure-functions-signalrservice-extension/compare/dev...Y-Sindo:pr/phantomcosmonaut/263?expand=1
I've put the attribute nested into `ServerlessHub` so that users can't use it outside of `ServerlessHub`. Also changed its name.

phantomcosmonaut:
Cool, thanks for looking into it. I think the attribute needs to be made public

src/SignalRServiceExtension/Config/OptionsSetup.cs
phantomcosmonaut, line 42: A common pattern for azure connection string settings

Y-Sindo: Ok.

Y-Sindo: > Cool, thanks for looking into it. Should the attribute just be `protected` and not `protected internal`?

It should be used in `SignalRTriggerBindingProvider` at https://github.com/Azure/azure-functions-signalrservice-extension/pull/263/files/077f73bbfb51e6b42f967170b8a1cabe713b16c3

Y-Sindo:
> Cool, thanks for looking into it. Should the attribute just be `protected` and not `protected internal`?

It should be used in `SignalRTriggerBindingProvider` at https://github.com/Azure/azure-functions-signalrservice-extension/pull/263/files/077f73bbfb51e6b42f967170b8a1cabe713b16c3

src/SignalRServiceExtension/ServerlessHubContextAttribute.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+namespace Microsoft.Azure.WebJobs.Extensions.SignalRService
+{
+  /// <summary>
+  /// Customized settings to be passed into the serverless hub context.
+  /// </summary>
+  [AttributeUsage(AttributeTargets.Class)]
+  class ServerlessHubContextAttribute : Attribute, IConnectionProvider
```
Y-Sindo, line 9: With `SignalRConnectionAttribute`, we can delete this class.

phantomcosmonaut: ah right, I should not have reverted that commit

phantomcosmonaut: done

phantomcosmonaut:
> > Cool, thanks for looking into it. Should the attribute just be `protected` and not `protected internal`?
> 
> It should be used in `SignalRTriggerBindingProvider` at https://github.com/Azure/azure-functions-signalrservice-extension/pull/263/files/077f73bbfb51e6b42f967170b8a1cabe713b16c3

Right, I was still using the old attribute. I have updated this

Y-Sindo (approved):

Issue #131, opened 2020-09-09
Support strongly typed SignalR Hubs


lfshr:
The `Hub` class allows for the use of a strongly typed Hub client. It would be great if the Azure SignalR service could also support the same! eg.
```
public class ClientHub : ServerlessHub<IChatClient>
{
    ...

    [FunctionName(nameof(SendClientMessage))]
    public async Task SendClientMessage([SignalRTrigger] InvocationContext invocationContext, ILogger logger)
    {
        await Clients.All.ReceiveMessage(...);
    }
}
```

https://docs.microsoft.com/en-us/aspnet/core/signalr/hubs?view=aspnetcore-3.1#strongly-typed-hubs

phantomcosmonaut:
I believe this is easily accomplished but I can't access the assembly `Microsoft.AspNetCore.SignalR.Hubs`.
```
using Microsoft.AspNetCore.SignalR;
using Microsoft.Azure.SignalR.Management;
using Microsoft.Azure.WebJobs.Extensions.SignalRService;

namespace foo
{
  public abstract class ServerlessHub<T> : ServerlessHub
      where T : class
  {
    protected ServerlessHub(IServiceHubContext hubContext = null, IServiceManager serviceManager = null)
      : base(hubContext, serviceManager)
    {
      Clients = new TypedHubConnectionContext<T>(base.Clients);
    }

    public new IHubClients<T> Clients { get; }
  }
}
```

phantomcosmonaut:
So... there are some hurdles casting between IHubClients<IClientProxy> and IHubClients<T> but here's the solution I came up with and it works because I just tested it. Commencing code barf 3, 2, 1

```
using Microsoft.Azure.SignalR.Management;
using Microsoft.Azure.WebJobs.Extensions.SignalRService;
using System;
using System.Collections.Generic;

namespace Microsoft.AspNetCore.SignalR
{
  // This is a frankenstein way to support typed client proxies in serverless hub class
  public abstract class ServerlessHub<T> : ServerlessHub where T : class
  {
    private IHubClients<T> _clients;
    public ServerlessHub(IServiceHubContext hubContext = null, IServiceManager serviceManager = null)
      : base(hubContext, serviceManager)
    {
    }
    public new IHubClients<T> Clients
    {
      get
      {
        var typedHub = typeof(IHubCallerClients).Assembly.GetType("Microsoft.AspNetCore.SignalR.Internal.TypedHubClients`1");
[...]

PR #264, opened 2021-08-21
Generic Serverless Hub for Typed Client Proxy

phantomcosmonaut:
Copied `TypedClientBuilder` code from [aspnetcore/signalr](https://github.com/dotnet/aspnetcore/blob/main/src/SignalR/server/Core/src/Internal/TypedClientBuilder.cs) and created a `TypedHubClients<T>` wrapper for `IHubClients<T>` that allows for a generic version of `ServerlessHub` accepting a custom  typed client proxy.

See https://github.com/Azure/azure-functions-signalrservice-extension/issues/131


Y-Sindo:
Thanks for your commit. I think it is better to implement `TypedClientProxy` from the [Azure SignalR Management SDK/Serverless SDK](https://github.com/Azure/azure-signalr/blob/dev/docs/management-sdk-guide.md) side so that we don't need to copy source code from ASP.NET CORE. Serverless SDK is the core dependency of SignalR extension. Actually we had a [PR](https://github.com/Azure/azure-signalr/pull/1034) on this, but at that time we were refactoring the whole serverless SDK, so we didn't accept it. We will put this issue into our backlog and we also welcome you to contribute on this.

phantomcosmonaut:
@Y-Sindo Ok, thanks for the feedback. I'll wait until I see this in the backlog

Y-Sindo:
Here is the issue in Serverless SDK: https://github.com/Azure/azure-signalr/issues/1033

PR #266, opened 2021-09-01
Fix thread-unsafe problem when creating `ServiceHubContext` and also simplify

Y-Sindo:


vwxyzh:
there is no thread issue.

src/SignalRServiceExtension/Config/ServiceHubContextStore.cs
```diff
@@ -28,36 +28,9 @@ public ServiceHubContextStore(IServiceEndpointManager endpointManager, IServiceM
...
-                store.TryUpdate(hubName, (null, value), pair);
-                return value;
-            }
-            catch (Exception)
-            {
-                store.TryRemove(hubName, out _);
-                throw;
-            }
+                new Lazy<Task<IServiceHubContext>>(
+                    () => ServiceManager.CreateHubContextAsync(hubName), true));
```
vwxyzh, line 32: will it throw exception?
do we want to cache exception?

PR #268, opened 2021-09-03
Move identity-based service endpoint parsing methods here

Y-Sindo:
- Move identity-based service endpoint parsing methods here
- Fix broken tests after SDK behaviour changes


JialinXin (approved):
:shipit:

